PR #150, opened 2026-04-11
fix: use local repoRes.data instead of stale repoData in issue count fallback

Grizouforever:
## Summary

In `RepositoryCheckTab.tsx`, the catch block for issue count fetching (line 95) references the `repoData` state variable. Inside the `useEffect` async callback, this is a stale closure value — `null` on first render since `setRepoData()` is async and the state hasn't updated yet. The fallback to `open_issues_count` never triggers.

## Changes

Replace `repoData` with the local `repoRes.data` variable which is already available in the same scope and holds the freshly fetched data.

The `eslint-disable-next-line react-hooks/exhaustive-deps` comment on line 110 is a further signal that dependency tracking was intentionally loosened, masking this stale reference.


Grizouforever:
Hi! This PR is ready for review — 37/37 tests pass. Happy to address any feedback or make adjustments. Thanks!

PR #176, opened 2026-04-13
fix: reset page to 0 when author filter chip is removed

### LINKED ISSUE #158: Author filter chip removal doesn't reset page to 0
## Description

In `src/components/miners/MinerPRsTable.tsx` (~line 249), the author filter chip's `onDelete` handler only calls `setSelectedAuthor(null)` but does **not** call `setPage(0)`:

```tsx
<Chip
  variant="filter"
  label={`Author: ${selectedAuthor}`}
  onDelete={() => setSelectedAuthor(null)}
/>
```

Every other filter control (status filter buttons: All, Open, Merged, Closed) correctly resets the page:

```tsx
onClick={() => {
  setStatusFilter('all');
  setPage(0);
}}
```

## Steps to Reproduce

1. Navigate to a miner's PR list with multiple pages.
2. Navigate to page 2 or beyond.
3. Type/select an author filter so fewer results show.
4. Delete the author filter chip.
5. The table may show an empty page because the page index is now out of range.

## Expected Behavior

Removing the author filter chip should reset the page to 0, consistent with all other filter controls.

## Actual Behavior

The page index is not reset, potentially leaving the user on an empty page.

TeapoyY:
### Problem
In MinerPRsTable.tsx, the author filter chip's `onDelete` handler only called `setSelectedAuthor(null)` but did **not** call `setPage(0)`. This caused the table to show an empty page when removing the author filter after navigating to a later page, because the filtered results no longer filled that page index.

### Fix
Added `setPage(0)` to the chip's `onDelete` handler, consistent with all other filter controls (status filter buttons) which also reset the page when changed.

### Changes
- src/components/miners/MinerPRsTable.tsx: updated onDelete handler to also reset page to 0

### Testing
- Navigate to page 2+ of a PR list
- Apply author filter
- Remove author filter chip
- Table now correctly shows page 1 (page index 0) instead of an empty page

Closes entrius/gittensor-ui#158

### DIFF
--- a/src/components/miners/MinerPRsTable.tsx
+++ b/src/components/miners/MinerPRsTable.tsx
@@ -246,7 +246,7 @@ const MinerPRsTable: React.FC<MinerPRsTableProps> = ({ githubId }) => {
               <Chip
                 variant="filter"
                 label={`Author: ${selectedAuthor}`}
-                onDelete={() => setSelectedAuthor(null)}
+                onDelete={() => { setSelectedAuthor(null); setPage(0); }}
               />
             )}
 


anderdc:
Closing in favor of #159 — same fix, #159 uses multi-line formatting which is more consistent with the rest of the codebase.

PR #214, opened 2026-04-14
refactor: Extract duplicated formatDate to shared utils

dataCenter430:
The `formatDate` function was copy-pasted identically across three issue components. Moved it to `src/utils/format.ts` and updated all consumers to import from the shared location.

- Removed duplicate from `IssueHeaderCard.tsx`
- Removed duplicate from `IssueSubmissionsTable.tsx`
- Removed duplicate from `IssuesList.tsx`

No functional changes.

### DIFF
--- a/src/components/issues/IssueHeaderCard.tsx
+++ b/src/components/issues/IssueHeaderCard.tsx
@@ -3,19 +3,9 @@ import { Box, Card, Typography, Chip, Link, Stack } from '@mui/material';
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import { IssueDetails } from '../../api/models/Issues';
 import { useStats } from '../../api';
-import { formatTokenAmount } from '../../utils/format';
+import { formatTokenAmount, formatDate } from '../../utils/format';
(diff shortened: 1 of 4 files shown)


dataCenter430:
Duplicated utility functions are a maintenance burden , a single fix would need to be applied in 3 places.

dataCenter430:
Hey, @anderdc 
Would be it possible to review my contributions soon?
please don't ignore them. 🙏 
Thanks

anderdc (changes requested):
Good dedup — three identical copies should be one.

One change: use `date-fns` `format()` instead of `toLocaleDateString`. PR #234 is consolidating the project onto date-fns and dropping dayjs, so this should align:

```ts
import { format } from 'date-fns';

export const formatDate = (dateStr: string | null | undefined): string => {
  if (!dateStr) return '-';
  return format(new Date(dateStr), 'MMM d, yyyy');
};
```

dataCenter430:
Just done! Clean and smart.
No conflict.
Would you please review the PR before some conflicts occur again? 😉 
Thank you.

anderdc (approved):

PR #215, opened 2026-04-14
refactor: Extract duplicated credibilityColor to shared utils

dataCenter430:
The `credibilityColor` function was copy-pasted identically across two miner score card components. Moved it to `src/utils/format.ts` and updated both consumers to import from the shared location.

- Removed duplicate from `MinerScoreCard.tsx` (also removed now-unused `CREDIBILITY_COLORS` theme import)
- Removed duplicate from `IssueDiscoveryScoreCard.tsx` (also removed now-unused `CREDIBILITY_COLORS` theme import)

No functional changes.


dataCenter430:
Having the same color logic in two components means any threshold change has to be made twice.

anderdc (approved):

PR #216, opened 2026-04-14
refactor: Extract repeated border and text color literals to theme constants

dataCenter430:
The string literals `rgba(255, 255, 255, 0.08)` and `rgba(255, 255, 255, 0.85)` were hardcoded across multiple unrelated components. Added `BORDER_SUBTLE` and `TEXT_MUTED` constants to `src/theme.ts` (alongside existing `RANK_COLORS`, `STATUS_COLORS`, etc.) and updated all affected files.

Files updated:
- `src/theme.ts` — added `BORDER_SUBTLE` and `TEXT_MUTED`
- `src/components/onboard/GettingStarted.tsx` — 3× border + 1× text
- `src/components/onboard/Scoring.tsx` — 2× border
- `src/pages/AboutPage.tsx` — 1× border
- `src/components/prs/PRDetailsCard.tsx` — 2× border
- `src/components/repositories/RepositoryScoreCard.tsx` — 1× border

No functional changes.

### DIFF
--- a/src/components/dashboard/LeaderboardCharts.tsx
+++ b/src/components/dashboard/LeaderboardCharts.tsx
@@ -13,6 +13,7 @@ import BarChartIcon from '@mui/icons-material/BarChart';
 import ReactECharts from 'echarts-for-react';
 import { useAllPrs, useReposAndWeights } from '../../api';
 import { truncateText } from '../../utils';
+import theme from '../../theme';
 
 const CHART_DOT_COLOR = 'rgba(88, 166, 255, 0.9)';
 
@@ -178,7 +179,7 @@ const LeaderboardCharts: React.FC = () => {
           margin: 12,
         },
         axisLine: {
-          lineStyle: { color: 'rgba(255, 255, 255, 0.08)', width: 1 },
+          lineStyle: { color: theme.palette.border.subtle, width: 1 },
         },
         axisTick: { show: false },
       },
@@ -201,7 +202,7 @@ const LeaderboardCharts: React.FC = () => {
         },
         splitLine: {
           lineStyle: {
-            color: 'rgba(255, 255, 255, 0.08)',
+            color: theme.palette.border.subtle,
             type: 'dashed',
             opacity: 0.5,
           },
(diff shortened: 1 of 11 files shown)


dataCenter430:
Magic strings make it hard to maintain consistency, a single constant change now updates all affected components.

anderdc (changes requested):
**Value conflict with existing palette** — `theme.ts` already defines `palette.border.subtle` as `rgba(255, 255, 255, 0.05)`, but the new `BORDER_SUBTLE` constant uses `rgba(255, 255, 255, 0.08)`. Two competing "subtle border" values in the same file with different opacities.

The fix should:
- Use `theme.palette.border.subtle` (already available via MUI's `sx` prop) and update its value to `0.08` if that's the intended one

Creating standalone exported constants bypasses the palette system the rest of the file uses.

**TEXT_MUTED naming is misleading** — `TEXT_MUTED` is `rgba(255, 255, 255, 0.85)` (85% opacity, near-full visibility). The existing `TEXT_OPACITY.muted` in the same file is `0.4`. Calling 85% "muted" contradicts the established scale. This is closer to `primary` than "muted."

**Incomplete migration** — PR updates 6 files but at least 4-5 more still have the same hardcoded `rgba(255, 255, 255, 0.08)`: `LeaderboardCharts.tsx`, `TopPRsTable.tsx`, `TopRepositoriesTable.tsx`, `RepositoryWeightsTable.tsx`, `LanguageWeightsTable.tsx`. If centralizing this value, the migration should be complete.

dataCenter430:
Thanks for your active review.
I will address it right now.

dataCenter430:
Maybe this approach is correct?
Removes standalone `BORDER_SUBTLE/TEXT_MUTED` exports in favor of the existing `palette.border.subtle` token (updated to 0.08). Completes the migration across all 11 files that had the hardcoded value.

Please review the PR again.
Thanks, @anderdc

dataCenter430:
Oh, one new conflict occurred for 1hr only, 😉 
I fixed it successfully.
Would you please review the PR once more? @anderdc 
I wonder if there will be another conflict again. 🙏

anderdc (approved):

PR #217, opened 2026-04-14
refactor: Extract duplicated FilterButton into shared component

dataCenter430:
The `FilterButton` component was defined identically (same styling, same props) inside three separate components. Extracted it to `src/components/FilterButton.tsx` and updated all consumers to import from the shared location.

- Removed local definition from `TopPRsTable.tsx`
- Removed local definition from `RepositoryPRsTable.tsx` (also removed unused `Button` import)
- Removed local definition from `RepositoryIssuesTable.tsx` (also removed unused `Button` import)
- Exported from `src/components/index.ts`

Call sites now pass `isActive` and `onClick` props instead of `value`, decoupling the component from internal state. No functional changes.

### DIFF
--- a/src/components/FilterButton.tsx
+++ b/src/components/FilterButton.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { Button } from '@mui/material';
+
+interface FilterButtonProps {
+  label: string;
+  isActive: boolean;
+  onClick: () => void;
+  count?: number;
+  color: string;
+  activeTextColor?: string;
+}
+
+const FilterButton: React.FC<FilterButtonProps> = ({
+  label,
+  isActive,
+  onClick,
+  count,
+  color,
+  activeTextColor = '#fff',
+}) => (
+  <Button
+    size="small"
+    onClick={onClick}
+    sx={{
+      color: isActive ? activeTextColor : 'rgba(255,255,255,0.5)',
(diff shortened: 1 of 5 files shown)


dataCenter430:
Copy-pasting a component across three files is a red flag ,any styling fix would need to be applied in three places. 🫤

dataCenter430:
Hi, @anderdc 
CI was already passed and there is no conflict now.
Ready to review.

anderdc (changes requested):
Good dedup — collapsing three identical definitions into one shared component with a clean `isActive`/`onClick` API is the right move.

Two things to fix since we're already refactoring this code:

- **Hardcoded rgba values in FilterButton.tsx** — replace with existing theme tokens: `rgba(255,255,255,0.5)` → `text.tertiary`, `rgba(255,255,255,0.1)` → `border.subtle`, `rgba(255,255,255,0.15)` → use `alpha()` or a border token. Same for the `#fff` default on `activeTextColor` → `text.primary`.
- **Inline `style` on the count span** — use `sx` instead. We want to minimize raw CSS in favor of MUI's sx prop.

dataCenter430:
Okay, thanks for your review.
I got it.

dataCenter430:
Addressed the review.
No conflict.

anderdc (approved):

PR #198, opened 2026-04-14
fix: persist PR list page number across navigation

leonaIee:
## Summary

Persist PR list page number in URL search params so navigating to a PR detail page and back preserves the current page instead of resetting to page 1.

Replaces `useState(0)` with `useSearchParams`-backed `prPage` parameter. Page 0 keeps the URL clean (no param); other pages set `prPage=N`. Browser back navigation restores the previous URL including the page param.

## Related Issues

Fixes #196

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Video

https://github.com/user-attachments/assets/7121ec98-f979-43d3-b79e-0d3b8042da7a

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #226, opened 2026-04-14
fix: persist PR list page state across navigation via URL params

### LINKED ISSUE #197: [BUG] Navigating via "Back to Repositories" button click from PR detail page resets Overview PR list table to first page.
## Description
This is about Overview tab PR list table.
Navigating via "Back to Repositories" button click from PR detail page resets PR list to the first page all the time.

## Steps to Reproduce
1. Visit https://gittensor.io/miners/details?githubId=124189514&tab=overview
2. Scroll down and jump to last page of PR list table (for example page number 8)
3. Click on a arrow icon(positioned at right side of the row) on a PR item, Click on "PR Details" link button(Blue color), You will be redirected to PR detail page.
 4. Click on "Back to Repositories" button, you are back to the page where PR list table shows.
5. Please see the PR list table page number became 1 though you were looking last page (for example 8)

## Expected
PR list table page number should be saved when you navigate back from PR detail page.

## Actual
PR list table page shows first page of long PRs list, user needs to go to last page again to check the last page PR items.
It's a broken experience for a user to check some PRs on the last page. As he have to scroll down, navigate to page 8, click a PR item again for several times.

## Note
This is probably keeping the page number as component local state instead of keeping as  URL param.

alexanderxfgl-bit:
## Summary
Fixes #197

## Problem
On the Overview tab, navigating to a PR detail page and clicking 'Back to Repositories' resets the PR list table to page 1, regardless of which page the user was viewing.

## Root Cause
The `page` state is stored in React `useState(0)`, which resets to 0 when the component unmounts during navigation to the PR detail page. When the user navigates back, the component remounts with page=0.

## Fix
Persist the page number in URL search params (`?prPage=N`) using `useSearchParams` from React Router:

- Page state initializes from URL param: `parseInt(searchParams.get('prPage') || '0')`
- Page changes update the URL via `updatePage()` helper with `replace: true` (no extra history entry)
- URL survives navigation — when user goes back, the page is restored from the URL
- When page resets to 0 (sort/filter changes), the param is removed from the URL

## Files Changed
- `src/components/miners/MinerPRsTable.tsx` — Added `useSearchParams`, `updatePage` helper, replaced all `setPage(0)` calls

### DIFF
--- a/src/components/miners/MinerPRsTable.tsx
+++ b/src/components/miners/MinerPRsTable.tsx
@@ -26,7 +26,7 @@ import {
   NavigateNext as NextIcon,
 } from '@mui/icons-material';
 import { useMinerPRs, type CommitLog } from '../../api';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 import ExplorerFilterButton from './ExplorerFilterButton';
 import { type MinerStatusFilter } from '../../utils/ExplorerUtils';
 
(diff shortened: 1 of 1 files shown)


anderdc:
closing in favor of #198 — same fix but #198 derives page directly from search params with no dual state or sync needed

Issue #196, opened 2026-04-13
[BUG] Navigating via "Back to Repositories" button click from PR detail page resets Pull Requests tab PR list to the first page all the time.


corevibe555:
## Description
This is about Pull Requests tab PR list table.
Navigating via "Back to Repositories" button click from PR detail page resets PR list to the first page all the time.

## Steps to Reproduce
1. Visit https://gittensor.io/miners/details?githubId=42954461&tab=pull-requests
2. Scroll down and jump to last page of PR list table (for example page number 8)
3. Click on a PR item on the table, you will be redirected to PR detail page.
4. Click on "Back to Repositories" button, you are back to the page where PR list table shows.
5. Please see the page number became 1 though you were looking last page (for example 8)

## Expected
PR list table page number should be saved when you navigate back from PR detail page.
So you should see the same page before you leaving to PR detail page.
The PR list table should be in visible area, the same as we were seeing before navigating to PR detail page.

## Actual
PR list table page shows first page of long PRs list, user needs to go to last page again to check the last page PR items.
The PR list table page is scrolled up to the top, so you have to scroll down again to see the PR table.

## Note
This is probably keeping the page number as component local state instead of keeping as  URL param.

PR #232, opened 2026-04-14
fix: persist Miner PR list pagination

### LINKED ISSUE #231: Fix: persist Miner PR list pagination across back navigation
## Description
On a miner's details page, the pull requests list forgets what page you were on whenever you open a PR and come back.

## Steps to Reproduce
1. Open any miner's details page, e.g. `/miners/details?githubId=124189514&tab=overview`
2. Scroll to the PR list and jump to a later page (say, page 8)
3. Click any PR to open its detail page
4. Click "Back to Repositories"

## Expected
The PR list is still on page 8 where I left it.

## Actual
The list has jumped back to page 1. I have to paginate all the way through again to find the PR I was looking at.

## Impact
This makes reviewing long PR histories tedious - the deeper into the list you are, the worse the regression. It breaks the mental model of "back means back".

eric-zachary077:
## Summary
Fixes #231 - the Miner Details --> Overview tab PR list reset to page 1 every time the user returned from a PR detail page. Now the current page is persisted in the URL and restored on back navigation.

## Root cause
`src/components/miners/MinerPRsTable.tsx` stored pagination as local `useState`. The `BackButton` on `PRDetailsPage` uses `navigate(-1)`, which restores the URL but not ephemeral component state, so every return remounted the table at page 0.

## Fix
Read/write the current page from a `prsPage` URL search param using `useSearchParams`, preserving all other params on the route (`githubId`, `tab`, etc.) via a functional update:

- Initialize `page` from `prsPage` query param on mount.
- On every `page` change, write it back to the URL with `{ replace: true }` so the history stack stays clean. `prsPage=0` is omitted to keep canonical URLs tidy.

The param name is namespaced (`prsPage`) to avoid colliding with any other table that may later live on the same miner details route.

No other state (search/sort/filter) is moved to the URL in this PR - the issue is scoped to page preservation and the user's reproduction specifically calls out page number.

## Type of Change
- [x] Bug fix

## Testing
- [ ] `npm run build`
- [ ] `npm run lint:fix`
- [ ] Manual (per issue repro):
  1. Open `/miners/details?githubId=124189514&tab=overview`
  2. Jump to the last page of the PR list (e.g. page 8)
  3. Click a PR row --> PR Details page
  4. Click "Back to Repositories"
  5. Verify the PR list is still on page 8 (URL now ends with `&prsPage=7` - zero-indexed)
- [ ] Direct URL load with `&prsPage=3` opens the table at page 4
- [ ] Refresh preserves page
- [ ] Page 0 doesn't add `prsPage=0` to the URL

## Checklist
- [x] No behavior changes outside pagination persistence
- [x] No new dependencies
- [x] Only `src/components/miners/MinerPRsTable.tsx` touched
- [x] Existing `githubId` / `tab` query params are preserved

## Notes
A sibling issue #196 reports the same bug for the Repository --> Pull Requests tab, which uses a different component (`RepositoryPRsTable`). Happy to follow up with a matching fix in a separate PR if this approach is accepted.

### DIFF
--- a/src/components/miners/MinerPRsTable.tsx
+++ b/src/components/miners/MinerPRsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useCallback } from 'react';
+import React, { useState, useMemo, useCallback, useEffect } from 'react';
 import {
   Card,
   Typography,
@@ -26,7 +26,7 @@ import {
   NavigateNext as NextIcon,
 } from '@mui/icons-material';
 import { useMinerPRs, type CommitLog } from '../../api';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 import ExplorerFilterButton from './ExplorerFilterButton';
 import { type MinerStatusFilter } from '../../utils/ExplorerUtils';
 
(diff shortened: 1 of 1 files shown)


johnstormentswe:
Hi @anderdc - this PR fixes by moving the PR list pagination state to the URL (`prsPage` query param), so the page is restored when the user returns from a PR detail via "Back to Repositories".

Scope is kept tight to one file (`MinerPRsTable.tsx`) and just the `page` state - search/sort/filter are untouched. Happy to follow up with the matching fix for #196 (Repository --> Pull Requests tab) in a separate PR if this approach looks good.

Please review when you have a moment - thanks!

anderdc:
contribution guidelines state an image is needed for ux changes, closing

johnstormentswe:
Hi @anderdc
I can understand why my PR was closed and really sorry not to know about this, 
I want you give me another change to upload an image so that I could contribute to this project.
Thanks.

PR #260, opened 2026-04-14
fix: normalize repo casing, refresh stale state, and stabilize repository PR row key

### LINKED ISSUE #258: Repository casing, stale React state, and unstable list key cause silent data loss and flicker
## Summary

Three small bugs surfaced across the miner and repository views. They all share the same flavour: data is technically present but silently fails to reach the UI correctly.

1. Repository name comparisons use strict equality or Map keys without casing normalization. Any drift between a stored `fullName` and an incoming `pr.repository` silently hides repos or mislabels them as untracked.
2. React state that should refresh when a dependency changes doesn't. The open-issues fallback in `RepositoryCheckTab` is dead code, and `MinerPRsTable`'s filters, search, and sort leak across miners during client-side navigation.
3. The repository PR table uses the array index in its row `key`, so sorting remounts every row instead of reordering in place.

Filing these together since the fixes are small, scoped to related areas, and naturally reviewed as one change.

### 1. Case-sensitive repository name lookups

GitHub treats repo names as case-insensitive, but several sites compare `fullName` with strict `===` or use it as a Map key without normalization. The codebase already has the correct pattern in `src/pages/search/searchData.ts` and the `repoStats` memo in `RepositoriesPage.tsx`, so the other sites just drifted.

* `src/pages/RepositoryDetailsPage.tsx`: `repos?.find((r) => r.fullName === repo)` flips the page into the "not tracked" empty state when the URL's `name` param has different casing than the stored `fullName`.
* `src/components/dashboard/LeaderboardCharts.tsx`: `repos.find((r) => r.fullName === repoName)` silently resolves the weight to `0` on a casing mismatch.
* `src/pages/RepositoriesPage.tsx`: both the `trendingRepos` and `topCollateralRepos` memos build `new Map(reposWithWeights.map((r) => [r.fullName, r]))` and check membership with `repoMap.has(pr.repository)`. Repos with casing drift are silently excluded from those widgets.

#### `RepositoryCheckTab`: open-issues fallback is dead code

**File:** `src/components/repositories/RepositoryCheckTab.tsx`

```ts
const repoRes = await axios.get(`.../repos/${repositoryFullName}`);
setRepoData(repoRes.data);                // schedules a re-render
// ...more awaits...
try {
  // search issues
} catch (e) {
  if (repoData && repoData.open_issues_count !== undefined) {
    setOpenIssuesCount(repoData.open_issues_count);
  }
}
```

The `repoData` inside the catch block is the **state variable from the closure**, not the freshly fetched response. React hasn't re-rendered between `setRepoData(...)` and the catch, so the closure still sees the initial `null` and the fallback never runs when the search-issues endpoint fails (rate limit, 422, network).

#### `MinerPRsTable`: filters leak across miners

**File:** `src/components/miners/MinerPRsTable.tsx`

Five pieces of local state (`selectedAuthor`, `statusFilter`, `searchQuery`, `sortField`, `sortDir`) are held as plain `useState` values with no tie to the `githubId` prop. Navigating from one miner to another via client-side routing carries the previous miner's filters into the new view.

### 3. Unstable row key in `RepositoryPRsTable`

**File:** `src/components/repositories/RepositoryPRsTable.tsx`

```tsx
{sortedPRs.map((pr, index) => (
  <TableRow key={`${pr.pullRequestNumber}-${index}`} ...>
```

The key includes the array `index`. Sorting reshuffles the array, so every row gets a new key on every sort and React remounts the entire table body (visible flicker plus wasted work).

## Expected behavior

* Lookups by repo name treat `Owner/Repo` and `owner/repo` as the same entity.
* `RepositoryCheckTab` falls back to the basic `/repos/...` response's `open_issues_count` when the search API fails.
* Each miner's PR table opens with defaults, with no carried-over filters, search, or sort.
* Sorting the repository PR table reorders rows in place.

## Actual behavior

* Repository detail page renders "not tracked" for a tracked repo on casing drift.
* Leaderboard chart shows `0` weight for active repos on casing drift.
* Trending and Most Collateral Staked widgets silently drop repos present in the main table.
* Open-issues count stays empty when the search API is rate-limited or errors out.
* Miner B's PR table opens with miner A's filters and sort.
* Sorting the repository PR table produces a visible flash as every row remounts.

## Steps to reproduce

1. Visit `/repositories?name=OWNER/REPO` where the stored `fullName` is `owner/repo`. The page shows the "not tracked" empty state.
2. On the Repositories page, compare the main table against the Trending and Most Collateral Staked widgets. Repos present in one but not the other indicate a casing drift.
3. Exhaust the unauthenticated GitHub search-issues rate limit (or intercept as 4xx / 5xx) and open a repository health-check tab. The open-issues count never populates, even though the basic `/repos/...` response returned `open_issues_count`.
4. On miner A's PR tab, apply an author filter, a status filter, type a search query, change the sort. Navigate to miner B client-side. Miner B's table opens with miner A's selections.
5. Open a repository PR table and click a sortable column header. Rows flash as they remount.

## Proposed fix

* Normalize both sides of every repo-name comparison with `.toLowerCase()` in `RepositoryDetailsPage.tsx`, `LeaderboardCharts.tsx`, and the two affected memos in `RepositoriesPage.tsx`.
* Read `repoRes.data` (local variable) in the catch block of `RepositoryCheckTab.tsx` instead of the stale `repoData` state.
* Add a `useEffect` keyed on `githubId` in `MinerPRsTable.tsx` that resets the five local filter / sort states.
* Replace the row key in `RepositoryPRsTable.tsx` with `${pr.repository}-${pr.pullRequestNumber}`.

Dexterity104:
## Summary

A handful of small, related fixes across the miner and repository views. They all share the same theme: data that was technically present but silently didn't reach the UI correctly.

* Repository name lookups are now case-insensitive, so casing drift between a stored `fullName` and an incoming `pr.repository` no longer hides repos from widgets or mislabels them as untracked.
* The open-issues fallback in `RepositoryCheckTab` now reads the local response variable instead of a stale state reference, so it actually runs when the search API fails.
* `MinerPRsTable` now resets its filter, search, and sort state when `githubId` changes, so one miner's selections no longer leak into the next.
* The repository PR table uses a stable row key, so sorting reorders rows in place instead of remounting the entire body.

Closes #258 

### Case-insensitive repository lookups

* `src/pages/RepositoryDetailsPage.tsx`: `.find((r) => r.fullName === repo)` now compares `r.fullName.toLowerCase() === repo?.toLowerCase()`. A mixed-casing URL no longer flips the page into the "not tracked" empty state.
* `src/components/dashboard/LeaderboardCharts.tsx`: same normalization inside the `.find(...)` that resolves repo weight, so weights no longer silently collapse to `0`.
* `src/pages/RepositoriesPage.tsx`: both the `trendingRepos` and `topCollateralRepos` memos build `repoMap` with `r.fullName.toLowerCase()` keys and look up with the lowercased `pr.repository`. Repos with casing drift are no longer excluded from those widgets.

The approach matches what `src/pages/search/searchData.ts` and the existing `repoStats` memo in `RepositoriesPage.tsx` already do.

### Stale React state

* `src/components/repositories/RepositoryCheckTab.tsx`: the catch block that runs when the search-issues requests fail read `repoData` from component state. Because `setRepoData(repoRes.data)` only schedules a re-render, the still-executing closure saw the initial `null` and the `if (repoData && ...)` branch never fired. Swapped the reference to `repoRes.data`, which holds the response captured from the first `await` in the same function. The fallback now populates `openIssuesCount` as intended.
* `src/components/miners/MinerPRsTable.tsx`: `selectedAuthor`, `statusFilter`, `searchQuery`, `sortField`, and `sortDir` were held as plain `useState` with no tie to `githubId`. Added a `useEffect` keyed on `githubId` that resets all five to their defaults when the prop changes. `useEffect` added to the React import.

### Stable row key in `RepositoryPRsTable`

* `src/components/repositories/RepositoryPRsTable.tsx`: row key was composed as `${pr.pullRequestNumber}-${index}`. Sorting reshuffled the array, so every row got a new key and React remounted the entire body. Swapped for `${pr.repository}-${pr.pullRequestNumber}`, which is unique across the dataset and stable under sort.

### DIFF
--- a/src/components/dashboard/LeaderboardCharts.tsx
+++ b/src/components/dashboard/LeaderboardCharts.tsx
@@ -56,7 +56,9 @@ const LeaderboardCharts: React.FC = () => {
     });
 
     statsMap.forEach((stats, repoName) => {
-      const repoData = repos.find((r) => r.fullName === repoName);
+      const repoData = repos.find(
+        (r) => r.fullName.toLowerCase() === repoName.toLowerCase(),
+      );
       if (repoData) {
         stats.weight = repoData.weight
           ? parseFloat(String(repoData.weight))
--- a/src/components/miners/MinerPRsTable.tsx
+++ b/src/components/miners/MinerPRsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useCallback } from 'react';
+import React, { useState, useMemo, useCallback, useEffect } from 'react';
 import {
   Card,
   Typography,
@@ -83,6 +83,14 @@ const MinerPRsTable: React.FC<MinerPRsTableProps> = ({ githubId }) => {
   const [searchQuery, setSearchQuery] = useState('');
   const [sortField, setSortField] = useState<PrSortField>('date');
   const [sortDir, setSortDir] = useState<SortDir>('desc');
+
+  useEffect(() => {
+    setSelectedAuthor(null);
+    setStatusFilter('all');
+    setSearchQuery('');
+    setSortField('date');
+    setSortDir('desc');
+  }, [githubId]);
 
   const page = parseInt(searchParams.get('prPage') || '0', 10);
   const setPage = useCallback(
--- a/src/components/repositories/RepositoryCheckTab.tsx
+++ b/src/components/repositories/RepositoryCheckTab.tsx
@@ -93,9 +93,9 @@ const RepositoryCheckTab: React.FC<RepositoryCheckTabProps> = ({
           setHelpWantedCount(hwRes.data.total_count);
         } catch (e) {
           console.warn('Failed to fetch issue counts', e);
-          // Fallback to repoData count if search fails, though it includes PRs
-          if (repoData && repoData.open_issues_count !== undefined) {
-            setOpenIssuesCount(repoData.open_issues_count);
+          // Fallback to repoRes.data (local variable) — repoData state is stale here
+          if (repoRes.data && repoRes.data.open_issues_count !== undefined) {
+            setOpenIssuesCount(repoRes.data.open_issues_count);
           }
         }
       } catch (err: unknown) {
--- a/src/components/repositories/RepositoryPRsTable.tsx
+++ b/src/components/repositories/RepositoryPRsTable.tsx
@@ -284,7 +284,7 @@ const RepositoryPRsTable: React.FC<RepositoryPRsTableProps> = ({
             <TableBody>
               {sortedPRs.map((pr, index) => (
                 <TableRow
-                  key={`${pr.pullRequestNumber}-${index}`}
+                  key={`${pr.repository}-${pr.pullRequestNumber}`}
                   onClick={() => {
                     navigate(
                       `/miners/pr?repo=${encodeURIComponent(pr.repository)}&number=${pr.pullRequestNumber}`,
--- a/src/pages/RepositoriesPage.tsx
+++ b/src/pages/RepositoriesPage.tsx
@@ -211,12 +211,16 @@ const RepositoriesPage: React.FC = () => {
       repoScores.set(pr.repository, cur);
     });
 
-    const repoMap = new Map(reposWithWeights.map((r) => [r.fullName, r]));
+    const repoMap = new Map(
+      reposWithWeights.map((r) => [r.fullName.toLowerCase(), r]),
+    );
 
     return Array.from(repoScores.entries())
       .filter(
         ([name, s]) =>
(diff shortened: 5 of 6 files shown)


anderdc:
Closing as duplicate — #261 addresses the same repo casing / stale state / unstable key issues and has CI passing. No need for two PRs fixing the same thing.

Dexterity104:
@anderdc 
I made that issue for this PR, I can't understand about this decision.

PR #261, opened 2026-04-14
fix: repo casing, react state, unstable list key

### LINKED ISSUE #258: Repository casing, stale React state, and unstable list key cause silent data loss and flicker

yunwei-zh:
## Summary

This PR fixes the bugs described in [#258](https://github.com/entrius/gittensor-ui/issues/258): GitHub-style case-insensitive repository identity across lookups and widgets, a stale React closure in the repository health-check tab, leaked miner PR table UI state on client-side navigation, and unstable table row keys causing unnecessary remounts when sorting.

Repository name casing

RepositoryDetailsPage: Resolving the tracked repo from useReposAndWeights() now compares fullName and the URL name query param case-insensitively, so a tracked repo is not shown as “not tracked” when casing differs (e.g. OWNER/REPO vs owner/repo).
LeaderboardCharts: Repo stats aggregate PRs by a normalized (lowercase) key and resolve weights with a case-insensitive match to the repos list; when a match exists, the displayed repository name uses the canonical fullName from the API.
RepositoriesPage: Trending, Most Collateral Staked, and Recent PRs use lowercase keys for internal maps and membership checks against reposWithWeights, and display names use the canonical fullName from the tracked repo map where applicable.
Stale state and keys

RepositoryCheckTab: When the GitHub Search API fails (rate limit, 4xx/5xx, etc.), the open-issues fallback now reads open_issues_count from the same repoRes.data returned by the basic /repos/{owner}/{repo} call, instead of the repoData React state variable (which was still stale in the catch closure).
MinerPRsTable: A useEffect keyed on githubId resets author filter, status filter, search query, sort field, and sort direction when switching miners so filters do not carry over via client-side routing.
RepositoryPRsTable: Table row key is ${pr.repository}-${pr.pullRequestNumber} instead of including the row index, so sorting reorders rows in place without remounting the whole body.

## Related Issues

fix #258 

## Type of Change

- Bug fix

## Testing

Manual checks recommended (test API per contributor workflow):
Repo details: Open /miners/repository?name=… with casing that differs from stored fullName; page should show tracked content, not the empty “not tracked” state.
Repositories widgets: Confirm Trending / Most Collateral / Recent PRs include repos that appear in the main table when PR repository strings differ only by case from fullName.
Leaderboard: Repo chart weights match tracked repos when PR repository casing differs from fullName.
Repository health tab: With Search API failing (or simulated), open-issues count still reflects open_issues_count from the basic repo response when available.
Miner PRs: On miner A, set filters/search/sort; navigate to miner B; table should reset to defaults.
Repository PRs: Sort columns; rows should update without a full flash/remount.

## Checklist

 New components are modularized/separated where sensible
 Uses predefined theme (e.g. no hardcoded colors)
 Responsive/mobile checked
 Tested against the test API
 npm run format and npm run lint:fix have been run
 npm run build passes
 Screenshots included for any UI/visual changes (N/A — no visual-only change)

### DIFF
--- a/src/components/dashboard/LeaderboardCharts.tsx
+++ b/src/components/dashboard/LeaderboardCharts.tsx
@@ -42,7 +42,8 @@ const LeaderboardCharts: React.FC = () => {
     const statsMap = new Map();
     allPRs.forEach((pr) => {
       if (!pr || !pr.repository) return;
-      const current = statsMap.get(pr.repository) || {
+      const repoKey = pr.repository.toLowerCase();
+      const current = statsMap.get(repoKey) || {
         repository: pr.repository,
         totalScore: 0,
         totalPRs: 0,
@@ -52,12 +53,13 @@ const LeaderboardCharts: React.FC = () => {
       current.totalScore += parseFloat(pr.score || '0');
       current.totalPRs += 1;
       if (pr.author) current.uniqueMiners.add(pr.author);
-      statsMap.set(pr.repository, current);
+      statsMap.set(repoKey, current);
     });
 
-    statsMap.forEach((stats, repoName) => {
-      const repoData = repos.find((r) => r.fullName === repoName);
+    statsMap.forEach((stats, repoKey) => {
+      const repoData = repos.find((r) => r.fullName.toLowerCase() === repoKey);
       if (repoData) {
+        stats.repository = repoData.fullName;
         stats.weight = repoData.weight
           ? parseFloat(String(repoData.weight))
           : 0;
--- a/src/components/miners/MinerPRsTable.tsx
+++ b/src/components/miners/MinerPRsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useCallback } from 'react';
+import React, { useState, useMemo, useCallback, useEffect } from 'react';
 import {
   Card,
   Typography,
@@ -83,6 +83,14 @@ const MinerPRsTable: React.FC<MinerPRsTableProps> = ({ githubId }) => {
   const [searchQuery, setSearchQuery] = useState('');
   const [sortField, setSortField] = useState<PrSortField>('date');
   const [sortDir, setSortDir] = useState<SortDir>('desc');
+
+  useEffect(() => {
+    setSelectedAuthor(null);
+    setStatusFilter('all');
+    setSearchQuery('');
+    setSortField('date');
+    setSortDir('desc');
+  }, [githubId]);
 
   const page = parseInt(searchParams.get('prPage') || '0', 10);
   const setPage = useCallback(
--- a/src/components/repositories/RepositoryCheckTab.tsx
+++ b/src/components/repositories/RepositoryCheckTab.tsx
@@ -93,9 +93,8 @@ const RepositoryCheckTab: React.FC<RepositoryCheckTabProps> = ({
           setHelpWantedCount(hwRes.data.total_count);
         } catch (e) {
           console.warn('Failed to fetch issue counts', e);
-          // Fallback to repoData count if search fails, though it includes PRs
-          if (repoData && repoData.open_issues_count !== undefined) {
-            setOpenIssuesCount(repoData.open_issues_count);
+          if (repoRes.data?.open_issues_count !== undefined) {
+            setOpenIssuesCount(repoRes.data.open_issues_count);
           }
         }
       } catch (err: unknown) {
@@ -109,7 +108,6 @@ const RepositoryCheckTab: React.FC<RepositoryCheckTabProps> = ({
     if (repositoryFullName) {
       fetchData();
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [repositoryFullName]);
 
   const checks: HealthCheck[] = useMemo(() => {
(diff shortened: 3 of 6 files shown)


anderdc (approved):
All four fixes are correct and address real bugs documented in #258:

- **Casing normalization**: `.toLowerCase()` on both sides of repo lookups — matches GitHub's case-insensitive behavior
- **Stale closure**: reading `repoRes.data` instead of the stale `repoData` state var in the catch block
- **Filter reset**: `useEffect` keyed on `githubId` to clear carried-over state between miners
- **Stable row key**: `${pr.repository}-${pr.pullRequestNumber}` instead of including array index

PR #263, opened 2026-04-14
refactor(pages): replace hardcoded colors with theme tokens

DragunovX16:
## Summary

- replace hardcoded page-level colors with existing theme tokens across Home, Dashboard, Issues, PR, Repository, Onboard, and Discoveries pages
- add `palette.alert.warningBg`, `palette.alert.warningBorder`, and `palette.alert.warningText` for warning alert styling
- update page tabs, borders, surfaces, scrollbar thumbs, and status chips to use semantic theme values instead of inline color literals

## Type of Change

- [x] Refactor
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Other (describe below)


anderdc (changes requested):
The page-level token swaps are good — those files should be using the theme tokens we already have. Two things to fix:

- **Drop the `palette.alert` addition from theme.ts.** Three tokens for one component isn't worth a new palette section. Use `alpha(STATUS_COLORS.warningOrange, 0.08)` and friends inline, or use the existing `status.warning` / `status.warningOrange` tokens directly.
- **`text.disabled` doesn't exist in the theme.** Used in 4 places in `RepositoriesPage.tsx`. The palette has `text.primary`, `text.secondary`, `text.tertiary` — pick the closest match or use `alpha(theme.palette.text.primary, 0.3)` to preserve the original appearance.

Also no CI checks ran on this branch — push or retrigger so build/format pass before re-review.

anderdc (changes requested):
Previous feedback on `palette.alert` and `text.disabled` is addressed — those are gone.

New issue: **this PR bundles two unrelated changes.** The title says "replace hardcoded page colors with theme tokens" but half the diff is a new `ensureArray` / `useArrayApiQuery` API safety layer across 7 API files. These should be separate PRs — the API changes touch shared data-fetching hooks and need their own review and testing.

Split this into:
1. Theme token swaps (pages only)
2. `ensureArray` / `useArrayApiQuery` API hardening (separate PR)

anderdc (approved):
API changes split out, pages-only now. CI passes. Looks good.

PR #281, opened 2026-04-15
fix: hide default webkit scrollbar button artifact in repo weights table

### LINKED ISSUE #280: Small visual artifact above the scrollbar on Onboard → Repositories tab
## Description
On the Onboard page (`/onboard?tab=repositories`), after selecting the Repositories tab and setting page size to 25 so the table becomes scrollable, a small dash appears just above the top of the scrollbar, inside the top-right corner of the table. It's a tiny artifact rather than actual content.

The same table on the other pages uses the same scrollbar styling, so the fix should apply wherever this pattern lives.

## Steps to Reproduce
1. Go to `https://gittensor.io/onboard?tab=repositories`
2. In the top-right of the table controls, set **Rows: 25**
3. Scroll the table
4. Look at the top-right corner of the table, right above the scrollbar
5. Notice a small dash/tick that shouldn't be there

## Expected
The scrollbar track should be clean all the way to the top of the table — no dash, tick, or extra pixel artifact above the scrollbar.

## Actual
A small dash is visible at the top of the scrollbar area.

## Screenshot
(attach the screenshot with the red circle)

## Environment
- Browser: Google Chrome (Webkit)
- URL: `https://gittensor.io/onboard?tab=repositories`

## Screenshots

Image

edwin-rivera-dev:
Fixes #280 
## Summary
Fixes a visual artifact on the Onboard → Repositories tab: when the table is scrollable (e.g. Rows ≥ 25), a thin line appears above the vertical scrollbar in the top-right corner.

## Root cause
In `src/components/repositories/RepositoryWeightsTable.tsx`, the `<Box>` wrapping the description + controls has a `borderBottom: '1px solid rgba(255,255,255,0.1)'`. That wrapper spans the full component width. The `TableContainer` beneath it is `overflow: auto`, so when the scrollbar appears it narrows the scrollable area by ~8px — but the border above doesn't shrink with it, leaving ~8px of border hanging over the top of the scrollbar track. That's the "dash" in the issue.

The sibling `LanguageWeightsTable` doesn't have this bug because it separates its controls from its table with `mb: 3` spacing instead of a full-width border, and the sticky header cells supply their own per-cell `borderBottom` that naturally stops where the scrollbar starts.

## Fix
Match the pattern used by `LanguageWeightsTable`:

- Remove `borderBottom: '1px solid rgba(255,255,255,0.1)'` from the outer header wrapper in `RepositoryWeightsTable.tsx`.
- Replace it with `mb: 3` for breathing room.

The sticky-header `borderBottom` on each `TableCell` already provides the visible separator between controls and rows, and it correctly shrinks with the scrollable area.

## Type of Change
- [x] Bug fix (visual)

## Testing
- [ ] `npm run build`
- [ ] `npm run lint:fix`
- [ ] Manual on `/onboard?tab=repositories`: set Rows to 25 → scrollbar appears → no dash above it.
- [ ] Manual on `/onboard?tab=languages`: unchanged (verifies the fix just aligns the two tabs).

## Checklist
- [x] No behavior change
- [x] No new dependencies
- [x] One file touched
- [x] Matches the pattern already used by `LanguageWeightsTable`

## Screenshots
image

### DIFF
--- a/src/components/repositories/RepositoryWeightsTable.tsx
+++ b/src/components/repositories/RepositoryWeightsTable.tsx
@@ -486,6 +486,14 @@ const RepositoryWeightsTable: React.FC = () => {
                 backgroundColor: 'rgba(255, 255, 255, 0.2)',
               },
             },
+            '&::-webkit-scrollbar-button': {
+              display: 'none',
+              height: 0,
+              width: 0,
+            },
+            '&::-webkit-scrollbar-corner': {
+              backgroundColor: 'transparent',
+            },
           }}
         >
           <Table stickyHeader>
--- a/src/components/repositories/RepositoryWeightsTable.tsx
+++ b/src/components/repositories/RepositoryWeightsTable.tsx
@@ -315,11 +315,7 @@ const RepositoryWeightsTable: React.FC = () => {
 
   return (
     <Box ref={containerRef}>
-      <Box
-        sx={{
-          borderBottom: '1px solid rgba(255, 255, 255, 0.1)',
-        }}
-      >
+      <Box sx={{ mb: 3 }}>
         <Box sx={{ p: 2, pb: 1 }}>
           <Typography variant="body2" color="text.secondary">
             Contribute to any of these projects to gain score and earn emissions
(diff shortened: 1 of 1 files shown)


edwin-rivera-dev:
@anderdc can you please review my PR and let me know your feedback?

anderdc (changes requested):
The description says this fixes a Chrome ::-webkit-scrollbar-button artifact, but the diff just swaps a borderBottom for a margin on the header wrapper Box. Can you clarify how removing a border between the header and the table addresses a scrollbar pseudo-element issue? The root cause described doesn't match the change.

edwin-rivera-dev:
Hi @anderdc 
You're right - the PR description is stale, my apologies.
I updated the PR description now so it reflects what the diff actually does.
Please review again and let me know asap. 
Thanks.

anderdc (approved):

PR #272, opened 2026-04-14
Redesign dashboard UI and align cards with theme styling

it-education-md:
## Summary

This PR redesigns the dashboard toward a cleaner command-center layout with clearer grouping, lighter card treatment, and better signal hierarchy.

## Related Issues

Closes #271

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

https://github.com/user-attachments/assets/3426224c-852e-4bac-952b-90d7aca09a8c

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #294, opened 2026-04-15
fix: live commit log animation, timer leak, and related bugs

### LINKED ISSUE #293: Live commit log animates every row on first load and leaks a setTimeout
## Description

The Live Commit Log on the dashboard has a few correctness issues that show up during normal use:

1. The "new entry" pulse animation fires on every commit when the log first loads, instead of only firing on commits that arrived after the page opened.
2. The 2-second `setTimeout` that clears the animation has no cleanup, so unmounting the dashboard while it is pending logs a state-update-on-unmounted-component warning and leaks the timer.
3. The infinite-scroll IntersectionObserver is torn down and recreated on every poll (every 10s) because `logEntries.length` is in the effect's dependency array.

While reviewing the same area, two related issues were also found:

4. `PRDetailsPage` calls `navigate()` during render when the URL is missing `repo` or `number`, which triggers a React warning and still fires the PR details request once with empty params.
5. `RepositoriesPage` builds `recentPrs` with `new Date(pr.mergedAt || new Date())`, which passes a `Date` object back into the `Date` constructor and loses millisecond precision.

## Steps to Reproduce

1. Open the dashboard with a fresh session and watch the commit log: every initial row pulses as if it were new.
2. Navigate away from the dashboard within 2 seconds of a new commit arriving and check the console.
3. Visit `/pr-details` without query params and check the console for the navigate-during-render warning.

## Expected Behavior

- Only commits that arrive after mount should pulse.
- No timer leaks or unmounted-state warnings.
- Missing-param routes should redirect cleanly.
- Date conversions should preserve precision.

## Actual Behavior

See description above.

Dexterity104:
## Summary

Fixes a handful of correctness issues in the dashboard and PR details flow:

- **LiveCommitLog**: extracted the stream-tracking logic into `useCommitLogStream`. The new hook uses an explicit first-fill guard (so initial rows no longer all pulse), tracks the animation timer in a ref with cleanup on unmount, classifies head updates by timestamp instead of list position, and caps the seen-IDs set so a long-lived dashboard cannot leak memory. Also dropped `logEntries.length` from the IntersectionObserver deps so the observer is no longer rebuilt on every poll.
- **PRDetailsPage**: replaced the `navigate()`-during-render with `<Navigate replace />`, and passed `enabled={hasParams}` to `usePullRequestDetails` so the request does not fire with empty params.
- **RepositoriesPage**: replaced `new Date(pr.mergedAt || new Date())` with a type predicate on the `.filter()` so `mergedAt` is narrowed to `string` and the fallback is no longer needed. Pulled the duplicated weight lookup into a small `repoWeight` helper.
- **DashboardApi**: included the resolved `requestUrl` and `limit` in the `useInfiniteCommitLog` query key so the cache stays correct across env changes.

## Related Issues

Closes #293

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

Not applicable; no visual changes.

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/api/DashboardApi.ts
+++ b/src/api/DashboardApi.ts
@@ -64,12 +64,12 @@ export const useInfiniteCommitLog = (options?: {
 }) => {
   const baseUrl = import.meta.env.VITE_REACT_APP_BASE_URL;
   const limit = 15;
+  const url = '/dash/commits';
+  const requestUrl = baseUrl ? `${baseUrl}${url}` : url;
 
   return useInfiniteQuery({
-    queryKey: ['useInfiniteCommitLog'],
+    queryKey: ['useInfiniteCommitLog', requestUrl, limit],
     queryFn: async ({ pageParam }: { pageParam: number }) => {
-      const url = '/dash/commits';
-      const requestUrl = baseUrl ? `${baseUrl}${url}` : url;
       const { data } = await axios.get<CommitLog[]>(requestUrl, {
         params: { page: pageParam, limit },
       });
--- a/src/components/dashboard/LiveCommitLog.tsx
+++ b/src/components/dashboard/LiveCommitLog.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef } from 'react';
 import {
   Card,
   CardContent,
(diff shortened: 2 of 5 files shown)


anderdc:
Closing — PR #272 moves and rewrites LiveCommitLog.tsx into a new dashboard structure. These changes would be overwritten when #272 lands.

Dexterity104:
@anderdc What do you mean exactly? Why are you rejecting all of my work?

PR #289, opened 2026-04-15
fix: show collateral score for open PRs and rename Merged column to Date

edwin-rivera-dev:
## Summary
Fixes #288 . Two small corrections to the Repository → Pull Requests table:

1. **Score column on OPEN PRs** now shows the PR's **collateral score** (already on `CommitLog` as `collateralScore`) styled in the open-status color and prefixed with `-` to signal it's a penalty. Open PRs no longer render a uniform `0.0000` column. Merged and closed PRs are unchanged - still show `score`.
2. The **"Merged"** header becomes **"Date"** and the cell now shows `mergedAt` for merged PRs and `prCreatedAt` otherwise. The column is no longer a sea of dashes on *All* / *Open* / *Closed* filters.

## Why
- Open PRs haven't been fully scored yet, so their `score` field is `0`. Showing `0.0000` gave miners and reviewers no actionable information. The collateral score is what's actively being applied to the miner while the PR sits open — it's the meaningful number.
- The Merged column was wasted whitespace for any PR state other than MERGED. Renaming to Date and backfilling with the creation date keeps the column useful across every filter.

## What changed
- `src/components/repositories/RepositoryPRsTable.tsx`:
  - Score cell now branches on `prState`: OPEN + non-zero `collateralScore` → display `-X.XXXX` in `status.open` color with a tooltip; otherwise display `score.toFixed(4)` in primary text color (current behavior).
  - Header cell renamed `Merged` → `Date`.
  - Date cell now uses `pr.mergedAt ?? pr.prCreatedAt`.

No new data fetched, no new dependencies. `CommitLog` already exposes `collateralScore` and `prCreatedAt`.

## Type of Change
- [x] Bug fix / UX correctness

## Testing
- [ ] `npm run build`
- [ ] `npm run lint:fix`
- [ ] Manual on any repository with a mix of PR states:
  - Score column: open PRs show `-{collateral}` in the open-status color with a tooltip; merged / closed rows show the normal score.
  - Open PRs with `collateralScore = 0` (rare edge case) fall back to showing `0.0000` rather than `-0.0000`.
  - Date column: merged rows show merge date; open / closed rows show creation date; no more dashes under normal data.
  - Hovering a negative collateral cell surfaces the tooltip "Collateral score applied while this PR is open".
- [ ] Mobile viewport: column width doesn't change meaningfully.

## Checklist
- [x] No new dependencies
- [x] One file touched
- [x] No API changes
- [x] Behavior unchanged for MERGED rows

## Video to upload

[...]

PR #295, opened 2026-04-15
fix: show collateral on open PRs and unify the Merged column into Date

### LINKED ISSUE #288: [UI] Pull Requests table: open PRs show "0" for Score, and Merged column is mostly empty
## Description
Two small problems on a repository's Pull Requests table make it less useful than it should be.

### 1. Every open PR shows a Score of `0.0000`
Open pull requests haven't finished scoring yet, so the Score column is effectively empty for every open row - it just reads `0.0000` all the way down. The more useful number for an open PR is its **collateral score** (the penalty that's currently being applied while the PR sits open). That value is tracked per PR but isn't visible anywhere on this page.

### 2. The "Merged" column is empty for most PRs
The last column is labelled **Merged** and shows a date - but only for merged PRs. Every open or closed row just shows a dash. On the *All*, *Open*, or *Closed* filters that's a column full of blanks.

Since the column is really just a date, it would be more useful to simply call it **Date** and fill it in for every PR.

## Steps to Reproduce
1. Open any repository with open PRs, for example:
   `https://gittensor.io/miners/repository?name=entrius/gittensor-ui`
2. Click the **Pull Requests** tab
3. Look at the Score column - every open PR shows `0.0000`
4. Look at the last column - it only has a date on merged rows, everything else is `-`

## Expected
- **Score column**: for open PRs, show the PR's collateral score instead of `0.0000`, styled so it's visually clear that it's a pending penalty (for example, a minus sign, or the same color as the OPEN status). For merged and closed PRs, keep showing the normal score.
- **"Merged" column**: rename it to **Date**. Show the merge date if the PR is merged, otherwise show when the PR was opened.

## Actual
- Score reads `0.0000` on every open PR
- "Merged" column is empty for anything not merged

## Why it matters
Right now, users looking at a repository's open PRs can't tell which ones are applying the biggest penalty to miners, and the date column only works for one of three PR states. Both fixes surface information that's already being tracked.

## Screenshot

Image

## Environment
- Browser: any
- URL: `https://gittensor.io/miners/repository?name=<any repo>` → Pull Requests tab

Dexterity104:
## Summary

Closes #288

The Pull Requests table on a repository page had two gaps that hid useful information:

1. Open PRs always showed a Score of `0.0000`, even when a collateral penalty was actively being applied.
2. The Merged column was empty for every non-merged PR, leaving Open and Closed rows with a lonely dash.

This change reworks the table so both columns stay informative across every PR state, and adds a few small affordances that make the penalty data actually usable.

## Changes

* **Score column for open PRs**: when an open PR has a non-zero collateral score, the cell now renders `−X.XXXX` in the open-status color with a tooltip explaining that the value is a penalty and will be replaced by the earned score once the PR is merged or closed.
* **Score color coding**: positive final scores render in the additions green, negative ones in the deletions red, zero in muted white. Makes scanning a long table much faster.
* **New Collateral column**: dedicated, sortable column so the penalty can be inspected and ranked on its own without conflating it with final score. Non-applicable rows show an em dash.
* **Renamed Merged to Date**: shows `mergedAt` for merged PRs and `prCreatedAt` for everything else, with a small `MERGED` / `CREATED` sub-label (color matched to the status palette) and a tooltip with the full timestamp.
* **Sortable headers**: PR #, Commits, +/-, Score, Collateral, and Date are all sortable via `TableSortLabel`. Score sort tie-breaks on collateral so the most-penalized open PRs still rise to the top within a score = 0 group.
* **Open collateral exposure summary**: small subtitle under "Pull Requests (N)" that totals the active collateral across all open PRs in the current filter, so the repo-wide cost is visible at a glance.
* **Header tooltips** on Score, Collateral, and Date so the dual meaning of each column is self-documenting.

## Files touched

* `src/components/repositories/RepositoryPRsTable.tsx`

No new data fetching, no new dependencies. `collateralScore` and `prCreatedAt` were already present on `CommitLog`.

[result-screen.webm](https://github.com/user-attachments/assets/490a822d-133d-43c8-a8b6-a16afa105ae9)

### DIFF
--- a/src/components/repositories/RepositoryPRsTable.tsx
+++ b/src/components/repositories/RepositoryPRsTable.tsx
@@ -9,10 +9,12 @@ import {
   TableContainer,
   TableHead,
   TableRow,
+  TableSortLabel,
   CircularProgress,
   Avatar,
   Chip,
   Stack,
+  Tooltip,
 } from '@mui/material';
 import { useAllPrs } from '../../api';
 import { useNavigate } from 'react-router-dom';
@@ -24,11 +26,22 @@ import {
   isOpenPr,
 } from '../../utils';
 import FilterButton from '../FilterButton';
+import type { CommitLog } from '../../api/models/Dashboard';
 
 interface RepositoryPRsTableProps {
   repositoryFullName: string;
   state?: 'open' | 'closed' | 'merged' | 'all';
 }
+
+type SortKey = 'pr' | 'commits' | 'changes' | 'score' | 'collateral' | 'date';
+type SortDir = 'asc' | 'desc';
+
+const num = (s?: string | null) => parseFloat(s || '0') || 0;
+
+const effectiveDateMs = (pr: CommitLog) => {
+  const iso = pr.mergedAt || pr.prCreatedAt;
+  return iso ? new Date(iso).getTime() : 0;
+};
 
 const RepositoryPRsTable: React.FC<RepositoryPRsTableProps> = ({
   repositoryFullName,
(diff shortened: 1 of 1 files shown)


anderdc:
Closing in favor of #289 which landed first and matches the scope of #288. The sortable columns and dedicated collateral column here go beyond what the issue requested.

Please put up a separate issue/PR for those features

PR #296, opened 2026-04-15
fix(security): prevent XSS by disabling raw HTML in markdown renderers

jsdevninja:
## Summary

This PR addresses a cross-site scripting (XSS) risk in markdown rendering across the UI.
Previously, several components used react-markdown with rehype-raw, which allows raw HTML from markdown input to be parsed into live DOM elements. Because these markdown sources come from external/untrusted content (GitHub repository docs and user-generated issue/PR text), this created a client-side injection path.

Why this was vulnerable

- rehype-raw explicitly enables raw HTML passthrough in markdown.
- External markdown (README / CONTRIBUTING / comments) can contain crafted HTML payloads.
- Rendering those payloads in-app can lead to script execution or malicious UI injection depending on browser behavior and payload structure.
- This is especially sensitive in issue/PR conversations where content is user-authored and high-volume.

## Behavioral impact

- Markdown is still rendered with remark-gfm, so standard GitHub-style markdown remains intact.
- Raw embedded HTML in markdown is now treated as plain text / not interpreted as active DOM.
- Existing safe behavior for external links (target="_blank" + rel="noopener noreferrer") remains unchanged.

## Security outcome

- Eliminates the direct raw-HTML rendering vector in the app’s markdown surfaces.
- Reduces attack surface in both repository document viewers and user-comment timelines.
- Improves defense-in-depth by removing the unneeded parser dependency entirely.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,6 @@
         "react-router-dom": "^6.25.1",
         "react-syntax-highlighter": "^16.1.0",
         "react-tooltip": "^5.30.0",
-        "rehype-raw": "^7.0.0",
         "remark-gfm": "^4.0.1"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


jsdevninja:
@LandynDev @anderdc Could you please rerun CI? Build is working on local.

anderdc:
Closing — React's DOM reconciler already prevents script execution from rehype-raw output, so the XSS vector described here doesn't exist in practice. Removing rehype-raw would break legitimate HTML rendering in READMEs (badges, HTML tables, details blocks) with no real security gain.

jsdevninja:
Hi, @anderdc 
Good point that React won't execute literal <script> tags from rehype-raw output.

But that does not make rehype-raw safe for untrusted markdown by itself. The unified/react-markdown guidance is to pair raw HTML parsing with sanitization (rehype-sanitize) when content is untrusted.

In our case, content is fetched from external repos and user-authored issue/PR text, so we should treat it as untrusted. The risk is broader than <script> tags (dangerous elements/attributes/protocols in raw HTML), and raw HTML bypasses markdown-only safety assumptions.

I agree we should preserve legitimate README HTML (badges/tables/details). A better middle ground is:

1.keep rehype-raw disabled for user-generated issue/PR comments, and
2. re-enable HTML for README/CONTRIBUTING with rehype-raw + rehype-sanitize allowlist schema.

That keeps feature parity while adding a real security boundary.

PR #298, opened 2026-04-15
fix: abort in-flight fetches in repo viewers to prevent stale content

### LINKED ISSUE #297: [BUG] Stale content from cancelled navigation can overwrite Repository viewers (README / CONTRIBUTING / Code tabs)
## Description
Three viewer components on the Repository Details page fetch remote content from jsDelivr via `axios` inside `useEffect`, but do not cancel the in-flight request when their input props change or the component unmounts. If a user switches repositories (or files, in the Code tab) faster than the network responds, a late response from the previous request overwrites state after the new navigation — the viewer ends up displaying content that does not match its current URL / header.

Affected files:
- `src/components/repositories/ReadmeViewer.tsx` — fetches `README.md` when `repositoryFullName` changes
- `src/components/repositories/ContributingViewer.tsx` — probes up to 6 branch/path combinations for `CONTRIBUTING.md` (worst in-flight window)
- `src/components/repositories/CodeViewer.tsx` — fetches raw content of any text file selected in the Code tab

## Steps to Reproduce
1. `npm run dev`, open the app.
2. DevTools → Network → throttle to **Slow 3G**.
3. Visit `/miners/repository?name=opentensor/bittensor`. A request to `cdn.jsdelivr.net/.../bittensor@main/README.md` appears as pending.
4. While that request is still pending, navigate to `/miners/repository?name=bitcoin/bitcoin`.
5. Wait for both requests to resolve.

**Most easily reproduced in the Code tab**: open the Code tab on any repo, click file A, immediately click file B. On slow networks, file A's content briefly displays under file B's header once A's request resolves.

## Expected Behavior
Changing `repositoryFullName` (or `filePath` in CodeViewer) should cancel any request started by the previous effect invocation. The viewer must only ever display content whose fetch was initiated by the current props.

## Actual Behavior
No `AbortController` and no cleanup function in any of the three `useEffect` hooks. Superseded requests continue to completion, and their `setContent(...)` call silently overwrites the viewer with stale data. The breadcrumb / URL / header still shows the new repo, but the body shows content from the previous repo.

## Impact
- **Data-integrity bug**: users can read documentation or source code from a different repository than the one they believe they are viewing. Misleading in context of the product (leaderboard / bounty / contribution decisions are made based on what the user sees here).
- Same root cause in three files → same fix pattern applied once per file.
- Silent on fast networks, which is why it hasn't surfaced earlier; reliably reproducible on throttled networks.

edwin-rivera-dev:
## Summary
Closes #297 . Prevents three viewer components on the Repository Details page from displaying stale remote content when the user navigates away mid-fetch.

## Root cause
`ReadmeViewer`, `ContributingViewer`, and `CodeViewer` each have a `useEffect` that calls `axios.get(...)` and writes the response into component state. None of them register a cleanup function to abort the in-flight request when the effect re-runs (e.g. `repositoryFullName` or `filePath` changes) or when the component unmounts. Late responses from superseded navigations still call `setContent`, so the viewer ends up showing content that does not match its current props.

`ContributingViewer` is the most exposed because it sequentially probes up to 6 branch/path combinations - a long in-flight window on slow networks — and writes partial state along the way.

## Fix
Each `useEffect` now creates an `AbortController`, passes `controller.signal` to every `axios.get` call, and calls `controller.abort()` from the cleanup function. `axios.isCancel` / `signal.aborted` checks prevent post-abort state writes. For `ContributingViewer`, the inner loop additionally checks `signal.aborted` before each request to avoid wasting network on an already-cancelled effect.

No changes to UX, markup, or request URLs. Axios has supported `{ signal }` natively since 0.22; the project is on a newer version, so no dependency bump is required.

## Type of Change
- [x] Bug fix (correctness)

## Testing
- [ ] `npm run build`
- [ ] `npm run lint:fix`
- [ ] Manual with DevTools Network throttled to "Slow 3G":
  - Open `/miners/repository?name=opentensor/bittensor`, then quickly navigate to `/miners/repository?name=bitcoin/bitcoin` before the README loads. Confirm the README tab only ever shows `bitcoin/bitcoin`'s README, never flips back to the first.
  - Same test for CONTRIBUTING tab (flip between repos faster than 6 fallback requests complete).
  - Same test for CodeViewer: click File A in the tree, immediately click File B. Confirm only File B's content is ever shown.
- [ ] Console is free of `CanceledError` traces (the catch blocks swallow them correctly).

## Checklist
- [x] No behavior change for legitimate users
- [x] No new dependencies
- [x] Three files touched
- [x] All state-writes guarded against stale responses

## Follow-ups (out of scope)
- The same anti-pattern exists in `RepositoryCheckTab.tsx`, `RepositoryCodeBrowser.tsx`, and `PRFilesChanged.tsx` (binary-file raw fetch). Separate PR.
- Longer-term, consider migrating these viewers to `useQuery` (TanStack Query is already used elsewhere in the codebase). That comes with automatic cancellation, caching, and dedup for free - but it's a rewrite, not a bug fix.

## Video to upload

https://github.com/user-attachments/assets/8296f83a-023a-42d7-a31c-a6a2619fb564

### DIFF
--- a/src/components/repositories/CodeViewer.tsx
+++ b/src/components/repositories/CodeViewer.tsx
@@ -29,6 +29,8 @@ const CodeViewer: React.FC<CodeViewerProps> = ({
     : '';
 
   useEffect(() => {
+    const controller = new AbortController();
+
     const fetchContent = async () => {
       if (!filePath || isImage) {
         // Don't fetch text content for images or if no file selected
@@ -43,19 +45,23 @@ const CodeViewer: React.FC<CodeViewerProps> = ({
         // Use raw.githubusercontent.com
         const response = await axios.get(rawUrl, {
           transformResponse: [(data) => data],
+          signal: controller.signal,
         }); // Force text
+        if (controller.signal.aborted) return;
         setContent(response.data);
       } catch (err) {
+        if (axios.isCancel(err) || controller.signal.aborted) return;
         console.error('Failed to fetch file content', err);
         setError(
           'Could not load file content. It might be binary or too large.',
         );
       } finally {
-        setLoading(false);
+        if (!controller.signal.aborted) setLoading(false);
       }
     };
 
     fetchContent();
+    return () => controller.abort();
   }, [repositoryFullName, filePath, defaultBranch, isImage, rawUrl]);
 
   if (!filePath) {
--- a/src/components/repositories/ContributingViewer.tsx
+++ b/src/components/repositories/ContributingViewer.tsx
@@ -20,6 +20,8 @@ const ContributingViewer: React.FC<ContributingViewerProps> = ({
   const [error, setError] = useState<string | null>(null);
 
   useEffect(() => {
+    const controller = new AbortController();
+
     const fetchContributing = async () => {
       setLoading(true);
       setError(null);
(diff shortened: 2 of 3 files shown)


edwin-rivera-dev:
@anderdc can you please review my PR and let me know your feedback?

anderdc (approved):

PR #299, opened 2026-04-15
fix: persist Top Miners filters in URL

jsdevninja:
### Motivation
Filter choices on Top Miners were lost on refresh and couldn’t be shared via URL, creating a frustrating user experience and making filtered views non-reproducible. This change makes filters persistent and shareable while fixing the flicker caused by URL/state sync conflicts.

This PR makes Top Miners filters URL-persistent and stable.

- Adds query-param sync for Top Miners filters (sort, eligible) in src/components/leaderboard/TopMinersTable.tsx.
- Restores selected filter state from the URL on load/refresh so filter choices are not lost.
- Keeps default URL clean by removing params when filters are at defaults.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

image
image

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/leaderboard/TopMinersTable.tsx
+++ b/src/components/leaderboard/TopMinersTable.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { Box, Typography, CircularProgress, Grid } from '@mui/material';
 import { alpha, type Theme } from '@mui/material/styles';
 import { useSearchParams } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


jsdevninja:
@LandynDev @anderdc Could you please rerun CI?

anderdc (dismissed):
**Simplify the cleanup effect — it's not needed.**

`getSortOptionFromQuery` already falls back to `'totalScore'` for invalid values, and `showEligibleOnly` only equals `true` when the param is exactly `"true"`. So invalid URL params are already handled on the read path. The `useEffect` that watches `searchParams` to strip invalid values adds ~20 lines for a case that's already covered — remove it.

**Minor:** `sortOption` via `useMemo` recomputes on any query param change (not just `sort`) since `searchParams` is a new object each time. Not a perf problem today, but you could just inline `getSortOptionFromQuery(searchParams.get(SORT_QUERY_PARAM), variant)` where needed or memoize on the string value instead of the object.

jsdevninja:
@anderdc Thank you for your review, I simplified it as you requested.

anderdc (changes requested):
Remove the cleanup useEffect — getSortOptionFromQuery already falls back to 'totalScore' for bad values and showEligibleOnly only matches "true" exactly. The read path handles invalid params, so the 20-line watcher is redundant.

Optional: sortOption useMemo recomputes on any query param change since searchParams is a new object each time. Could memoize on the string value instead.

anderdc (approved):

PR #300, opened 2026-04-15
fix: cancel stale repo viewer fetches so content matches current props

### LINKED ISSUE #297: [BUG] Stale content from cancelled navigation can overwrite Repository viewers (README / CONTRIBUTING / Code tabs)

fansilas:
## Summary

Three viewers on the Repository Details page (`ReadmeViewer`, `ContributingViewer`, `CodeViewer`) fetched content in `useEffect` with no cleanup. On slow networks, switching repos or files before a request resolved let the old response land and overwrite the new view with stale content.
`ContributingViewer` was the worst offender since it probes up to six branch/path combinations per load.

## Approach

The same cancellation pattern would repeat in every viewer, so the fix adds two small shared primitives instead of inlining the same boilerplate three times:

- **`src/hooks/useAbortableEffect.ts`** owns the `AbortController` lifecycle and hands the effect a signal. Cleanup is automatic, so it cannot be forgotten in future viewers.
- **`src/components/repositories/jsdelivrFetch.ts`** exposes `fetchFirstMarkdown`, collapsing the branch/path probe loop that
  `ReadmeViewer` and `ContributingViewer` were each open coding. It is abort aware and returns `null` on cancellation.

Each viewer now passes the signal to `axios.get` and guards state writes with `signal.aborted`. URLs, fallback order, loading states, and error messages are unchanged.

## Related Issues

Fixes #297

## Type of Change

- [x] Bug fix
- [x] Refactor

## Screenshots

N/A, no visual change.

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/repositories/CodeViewer.tsx
+++ b/src/components/repositories/CodeViewer.tsx
@@ -1,10 +1,14 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { scrollbarSx } from '../../theme';
 import { Box, Typography, CircularProgress, Alert } from '@mui/material';
 import axios from 'axios';
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { vscDarkPlus } from 'react-syntax-highlighter/dist/esm/styles/prism';
 import ReactMarkdown from 'react-markdown';
+import {
+  isAbortError,
+  useAbortableEffect,
+} from '../../hooks/useAbortableEffect';
 
 interface CodeViewerProps {
   repositoryFullName: string;
@@ -29,8 +33,8 @@ const CodeViewer: React.FC<CodeViewerProps> = ({
     ? `https://cdn.jsdelivr.net/gh/${repositoryFullName}@${defaultBranch}/${filePath}`
     : '';
 
-  useEffect(() => {
-    const fetchContent = async () => {
+  useAbortableEffect(
+    async (signal) => {
       if (!filePath || isImage) {
         // Don't fetch text content for images or if no file selected
         setContent(null);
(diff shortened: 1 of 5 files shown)


anderdc:
Closing — duplicate of #298 which addresses the same stale-fetch bug with a lighter approach.

fansilas:
@anderdc 
Thanks for the look, but I don't think this is quite a duplicate.

Same bug, different shape. The other PR copies the same abort setup into all 3 viewers. This one pulls it into a small `useAbortableEffect` hook, and a `fetchFirstMarkdown` helper also cleans up the branch/path probing that `ReadmeViewer` and `ContributingViewer` were each doing by hand.

That matters for the next 3 files the issue already flags (`RepositoryCheckTab`, `RepositoryCodeBrowser`, `PRFilesChanged`), since each one becomes a one-line change with the hook in place instead of another copy of the same boilerplate.

Would you mind taking another look before closing it?

PR #317, opened 2026-04-15
fix(api): normalize list query responses to arrays

DragunovX16:
## Summary

- normalize list-style API query responses to arrays before exposing them to UI consumers
- harden shared list query hooks used by search, dashboard, repositories, miners, and issues views
- prevent render-time crashes when the test API returns a non-array payload for list endpoints

## Related Issues

Follow-up to the page-only theme token refactor PR (#263).
Addresses runtime crashes such as:
- `miners.map is not a function`
- `allPRs.forEach is not a function`

## Type of Change

- [x] Bug fix
- [ ] Refactor
- [ ] New feature
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] Responsive/mobile checked where relevant
- [x] `npm run format:check` has been run
- [x] `npm run lint` has been run
- [x] `npm run build` passes

### DIFF
--- a/src/api/ApiUtils.ts
+++ b/src/api/ApiUtils.ts
@@ -1,6 +1,9 @@
 import { useQuery } from '@tanstack/react-query';
 import axios, { type AxiosError } from 'axios';
 
+export const ensureArray = <T>(value: unknown): T[] =>
+  Array.isArray(value) ? value : [];
+
 export const useApiQuery = <TResponse = void, TSelect = TResponse>(
   queryName: string,
   url: string,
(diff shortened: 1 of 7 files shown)


anderdc:
Closing — the backend list endpoints already return arrays; guarding every consumer with ensureArray adds boilerplate for a scenario that doesn't occur in practice. If a specific endpoint is misbehaving, that should be fixed at the API layer.

9876543210-tc-0123456789:
Hi, @anderdc 

I rechecked this on a clean local checkout of the current `test` branch, and there is still a reproducible runtime crash without the follow-up guard work.

Repro:
1. checkout latest `test`
2. run `npm run dev`
3. open the app locally

Observed error:
- `miners.map is not a function`
- thrown from `GlobalSearchBar` via `useSearchResults` / `buildMinerSearchData`

So I agree the previous `ensureArray` / `useArrayApiQuery` PR was too broad and not the right abstraction for every list hook. But some fix still seems required, because the current frontend on `test` is crashing against the current runtime payload it receives.

I can rework this into a much narrower follow-up PR limited to the specific failing path instead of normalizing every list endpoint globally. If you prefer this to be fixed strictly at the backend/API contract layer instead, I’m fine with that too, but I wanted to call out that the crash is currently reproducible on `test`. 😅

PR #332, opened 2026-04-15
fix: remove N+1 PR details requests from LiveCommitLog

edwin-rivera-dev:
## Summary
Removes the unnecessary per-entry `usePullRequestDetails` fetch from the Dashboard live activity feed. Eliminates N simultaneous `GET /prs/details` requests on initial load (15 with the default page size) and every 10-second auto-refresh, with no change to what the user sees.

## Root cause
`CommitLogItem` is rendered once per entry in the live feed. Inside its body it unconditionally called:

```tsx
const { data: details } = usePullRequestDetails(
  entry.repository,
  entry.pullRequestNumber,
);

### DIFF
--- a/src/pages/dashboard/views/LiveCommitLog.tsx
+++ b/src/pages/dashboard/views/LiveCommitLog.tsx
@@ -12,7 +12,7 @@ import {
   Chip,
 } from '@mui/material';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


edwin-rivera-dev:
Hi @anderdc , Could you please review my PR and let me know asap?

anderdc (approved):

PR #342, opened 2026-04-15
refactor: extract MONO_FONT constant to theme.ts and replace all hardcoded font strings

dataCenter430:
`"'JetBrains Mono"', monospace"` was hardcoded in 54 files. This adds `MONO_FONT` to `theme.ts` alongside other shared constants (`STATUS_COLORS`, `scrollbarSx`, etc.), updates `leaderboard/types.ts` to derive `FONTS.mono` from it, and replaces all occurrences. No behavior change.

### DIFF
--- a/src/components/ContributionHeatmap.tsx
+++ b/src/components/ContributionHeatmap.tsx
@@ -41,7 +41,6 @@ const ContributionHeatmap: React.FC<ContributionHeatmapProps> = ({
         <Typography
           sx={{
             color: 'text.primary',
-            fontFamily: '"JetBrains Mono", monospace',
             fontWeight: 700,
             fontSize: '2.5rem',
             lineHeight: 1,
@@ -53,7 +52,6 @@ const ContributionHeatmap: React.FC<ContributionHeatmapProps> = ({
           variant="body2"
           sx={{
             color: alpha(theme.palette.common.white, TEXT_OPACITY.faint),
-            fontFamily: '"JetBrains Mono", monospace',
             fontSize: '0.85rem',
             mt: 0.5,
           }}
@@ -77,7 +75,6 @@ const ContributionHeatmap: React.FC<ContributionHeatmapProps> = ({
             <Typography
               sx={{
                 color: alpha(theme.palette.common.white, TEXT_OPACITY.muted),
-                fontFamily: '"JetBrains Mono", monospace',
                 fontSize: '0.85rem',
                 textAlign: 'center',
               }}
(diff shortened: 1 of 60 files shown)


anderdc (changes requested):
The theme already defines typography variants that use JetBrains Mono (`mono`, `monoSmall`, `dataValue`, `dataLabel`, `sectionTitle`, `tableHeader`, `statValue`, `statLabel`, `tooltipLabel`, `tooltipDesc`). Use those instead of a new constant — they're the actual "define it in one spot" mechanism.

- **`<Typography>` components**: use the variant (`variant="monoSmall"`) instead of `sx={{ fontFamily: MONO_FONT }}`
- **Non-Typography elements**: spread the theme typography (`sx={{ ...theme.typography.mono }}`)
- **MUI component defaults** (e.g. all TableCells): set it once in theme `components` overrides

`MONO_FONT` just moves the string into a variable — it's still repeated 54 times. The variants already exist and already solve this.

dataCenter430:
Hi, @anderdc 
Thanks for your detailed review.
I've fixed all as you wanted.
Ready to review.

anderdc (changes requested):
The codebase has 302 usages of JetBrains Mono vs 11 of Inter (all 11 are in theme.ts itself — no component actually uses Inter). The entire app is a monospace dashboard UI.

Rather than extracting a MONO_FONT constant that still needs to be specified on every component, the right fix is making JetBrains Mono the default font in the theme. That eliminates ~300 inline fontFamily declarations in one shot instead of replacing them with a constant.

The few Inter declarations in theme.ts (h1-h6, body1, body2, button) can be removed or switched to JetBrains Mono since nothing relies on them being Inter.

dataCenter430:
Good point, @anderdc  👍 
 makes much more sense to set it as the theme default than to push it onto every component. 
I'll update the PR to set `typography.fontFamily` to `JetBrains Mono`, remove the Inter overrides on h1–h6/body1/body2/button, and strip all the per-component `fontFamily` declarations. The semantic variants (`monoSmall, dataLabel, etc`.) will keep their other properties but lose the redundant fontFamily field.

dataCenter430:
Hi, @anderdc 
I've fixed them again as you wanted.
Would be it possible to review the PR again?
Thanks

anderdc:
sorry many conflicts, lots of moving PRs!

dataCenter430:
Yup, I knew.
Now I'm fixing them.
It's related to recent PRs merged.
Just one a sec. thanks

dataCenter430:
Hi, @anderdc 
I've resolve the conflicts.
Would you please review the PR again?

anderdc:
can I get one image to verify this PR does not affect any of the current styles/typography fonts

dataCenter430:
> can I get one image to verify this PR does not affect any of the current styles/typography fonts

Yup, Sure

dataCenter430:
image
image
image

dataCenter430:
There is no affects with my changes,
Anyway, hope your review.
Thanks

anderdc (approved):

PR #349, opened 2026-04-16
refactor: consolidate loading/error/empty states via QueryBoundary (#347)

### LINKED ISSUE #347: Refactor: consolidate loading/error/empty states via QueryBoundary
## Summary

Data-driven pages across the app handle the three states of a TanStack Query result — `isLoading`, `isError`, and `data` — in ad-hoc ways that have drifted between files. This issue proposes a shared `<QueryBoundary>` component to standardize that handling, so every page renders loading, empty, and error states consistently.

## Motivation

Today each page reimplements its own variation of the same three-state flow. The pattern is reasonable on a per-page basis, but has diverged enough that consistency is no longer free, and `isError` is rarely handled explicitly.

JPette1783:
## Summary

Introduces a shared `QueryBoundary` component that renders the four states of a TanStack Query result — loading / error / empty / data — with consistent default UI and per-call-site overrides. Migrates `PRDetailsPage` and `IssueDetailsPage` onto it so that error handling is explicit instead of falling through to the `!data` branch.

### Changes

- **New:** `src/components/common/QueryBoundary.tsx` — generic component with default `loading` / `error` / `empty` renderers, all overridable via `renderLoading`, `renderError`, `renderEmpty`, and an optional `isEmpty` predicate. Exposes a pure `selectQueryBranch` helper for unit testing.
- **New:** `src/tests/QueryBoundary.test.ts` — 9 unit tests covering branch precedence (`loading > error > empty > data`), the `isEmpty` predicate, null/undefined handling, and falsy-but-defined data (`0`, `''`, `false` stay on the `data` branch).
- **Migrated:** `src/pages/PRDetailsPage.tsx` — the `isLoading` / `!prDetails` cascade is replaced with `<QueryBoundary>`. `renderEmpty` and `renderError` both use the existing "PR not found" UI, so user-visible output is unchanged.
- **Migrated:** `src/pages/IssueDetailsPage.tsx` — same pattern. SEO tags now read from `issueQuery.data?.*` so the document title still populates on success.
- **Exported:** `QueryBoundary` from `src/components/common/index.ts`, reachable via the existing barrel in `src/components/index.ts`.

### Behavior

No user-visible change intended. Loading spinners, "not found" copy, and back-button placements render identically to before. The primary upgrade is structural: `isError` is now handled explicitly by the component, whereas previously a network failure fell through to `!data` and was indistinguishable from a real "not found" response. Both are mapped to the same UI in this PR to preserve existing copy; future PRs can surface a distinct "network error" message without touching either page.

## Related Issues

Closes #347 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

Not applicable — no UI/visual changes. Loading state, "not found" state, and rendered page content are byte-identical to the prior implementation.

## Checklist

- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes *(N/A — no visual changes)*

### DIFF
--- a/src/components/common/QueryBoundary.tsx
+++ b/src/components/common/QueryBoundary.tsx
@@ -1,0 +1,99 @@
+import React from 'react';
+import { Alert, Box, CircularProgress, Typography } from '@mui/material';
+import { type UseQueryResult } from '@tanstack/react-query';
+
+export type QueryBranch = 'loading' | 'error' | 'empty' | 'data';
+
+type MinimalQueryResult<T> = {
+  data: T | undefined;
+  isLoading: boolean;
+  isError: boolean;
+};
+
+// Pure branch selector, exposed for unit testing. Determines which of the four
+// UI branches a query result should render, using the same precedence the
+// component uses: loading > error > empty > data.
+export const selectQueryBranch = <T,>(
+  query: MinimalQueryResult<T>,
+  isEmpty?: (data: T) => boolean,
+): QueryBranch => {
+  if (query.isLoading) return 'loading';
+  if (query.isError) return 'error';
+  if (query.data === undefined || query.data === null) return 'empty';
+  if (isEmpty && isEmpty(query.data)) return 'empty';
+  return 'data';
+};
+
+type QueryBoundaryProps<T> = {
+  query: UseQueryResult<T>;
+  children: (data: T) => React.ReactNode;
+  renderLoading?: () => React.ReactNode;
+  renderError?: (error: unknown) => React.ReactNode;
+  renderEmpty?: () => React.ReactNode;
+  isEmpty?: (data: T) => boolean;
+};
+
+const centeredBoxSx = {
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  minHeight: '50vh',
+  flexDirection: 'column' as const,
+  gap: 2,
+};
+
+const DefaultLoading: React.FC = () => (
+  <Box sx={centeredBoxSx}>
+    <CircularProgress />
+  </Box>
+);
+
+const DefaultError: React.FC<{ error: unknown }> = ({ error }) => (
+  <Box sx={centeredBoxSx}>
+    <Alert severity="error">
+      {error instanceof Error && error.message
+        ? error.message
+        : 'Something went wrong while loading this page.'}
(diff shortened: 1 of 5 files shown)


anderdc (changes requested):
Reasonable refactor, component is well-typed and tests are solid. One issue:

**SEO tags break on loading/error states** — In `IssueDetailsPage`, the SEO tags now read from `issueQuery.data?.title` etc. Before the refactor, `issue` was set from `data` which was only accessed inside the `else` (data) branch, so SEO always had the fallback on loading. That part is fine. But the real problem is that `<SEO>` is rendered **outside** `<QueryBoundary>`, so it renders on every state with the optional-chained fallbacks. This is actually the same behavior as before — just flagging that the SEO placement was already imperfect and this refactor didn't address it.

Actually, on closer look the behavior is identical to before. **Withdraw this concern — approve.**

Minor nits (non-blocking):
- The `renderNotFound` box in both pages duplicates `centeredBoxSx` from QueryBoundary itself. Could expose that style or let the default renderers handle it with custom message props instead of full render overrides.
- Only 2 consumers so far. The abstraction pays for itself if more pages adopt it, but watch that it doesn't become a premature pattern that nobody else picks up.

Looks good overall.

anderdc (approved):
Looks good. Component is well-typed, tests cover the branch precedence correctly, and the migration preserves existing behavior.

Minor nits (non-blocking):
- `renderNotFound` in both pages duplicates `centeredBoxSx` from QueryBoundary. Could expose that style or add a `message` prop to the default empty/error renderers to avoid the full render overrides.
- Only 2 consumers — the abstraction pays for itself if more pages adopt it.

anderdc:
Closing this — the abstraction doesn't earn its weight here.

- Both consumers override `renderEmpty` and `renderError`, so the default renderers are dead code from day one.
- The explicit `isError` handling (the main claimed improvement) is solved by adding one `if (isError)` line to each page — no new component needed.
- The two pages have different "not found" UI ("PR not found" vs "Issue not found" with different back buttons), so the pattern isn't actually shared, just the skeleton.
- Adds indirection for no payoff — a reader now has to jump to `QueryBoundary.tsx` to understand flow that was previously obvious inline.

PR #350, opened 2026-04-16
refactor: standardize numeric parsing in MinerScoreCard

jerryzai:
## Summary
- Replace raw `Number(x || 0)` coercions in `MinerScoreCard` StatTile display values with `parseNumber(x)` from `ExplorerUtils`.
- `parseNumber` guards with `Number.isFinite`, so `NaN`/`Infinity`/`null`/`undefined` fall back to `0` instead of surfacing as `"NaN"` in the UI.
- Scope limited to display coercions (lines 591, 612, 613, 621, 657, 664, 666, 673). Sort comparators in `rank()` are left untouched.

## Why
`parseNumber` was already imported and used elsewhere in this file (lines 356–358) for `credibility`, `totalOpenPrs`, `totalCollateralScore`. This finishes the standardization so the display values use the same safe helper and the `|| 0` / `?? 0` fallbacks become redundant.

## Test plan
- [ ] Open a miner detail page in PRs view — Score, Token Score, PRs sub-line render the same values as before
- [ ] Open a miner detail page in Discoveries view — Score, Credibility, Token Score render the same values
- [ ] Verify a miner with missing fields no longer shows `NaN` in any StatTile


anderdc:
Closing — this is the same pattern as #352 (same author, same change, adjacent file). Please combine both into a single PR so we're not reviewing the same mechanical find-and-replace twice. The two files don't overlap so it should be a clean combine.

PR #358, opened 2026-04-16
refactor: remove dead useCommitLog hook

wdeveloper16:
### Summary
- Removes `useCommitLog` which has been superseded by `useInfiniteCommitLog` and is not imported anywhere in the codebase.

### Why it's safe to remove
- No imports of `useCommitLog` exist outside its own definition.
- `useInfiniteCommitLog` in the same file covers the same endpoint (`/dash/commits`) with pagination support.
- The `CommitLog` type import is retained — it is still used by `useInfiniteCommitLog`.

### Changes
- `src/api/DashboardApi.ts` — deleted 12 lines (the `useCommitLog` function)

### DIFF
--- a/src/api/DashboardApi.ts
+++ b/src/api/DashboardApi.ts
@@ -47,18 +47,6 @@ export const useReposAndWeights = () =>
 export const useLanguagesAndWeights = () =>
   useDashboardQuery<LanguageWeight[]>('useLanguagesAndWeights', '/languages');
 
-export const useCommitLog = (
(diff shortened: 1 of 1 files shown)


anderdc:
Closing — please combine this with #360 and #370 into a single dead-code cleanup PR. They're all the same type of change (removing unused hooks/exports) and don't need to be separate PRs. Keep #359 as the base since it already has the changes, and fold the others in.

PR #359, opened 2026-04-16
Refactor/remove use issue

wdeveloper16:
### Summary
- Removes three unrelated dead exports in a single cleanup PR — all confirmed unused across the full codebase including tests.

### Why it's safe to remove
- `useIssue`: never imported anywhere; `IssueDetailsPage` uses `useIssueDetails` instead (different endpoint). `IssueBounty` type import is retained — still used by `useIssues` and `useRepoIssues`.
- `useRepositoryPRs`: never imported anywhere. Also cleans up the now-orphaned `CommitLog` import in the same file.
- `routePaths`: never imported anywhere. `getRouteForPathname` (retained) covers the same lookup pattern.

### Changes
- `src/api/IssuesApi.ts` — deleted 12 lines (`useIssue` function + JSDoc)
- `src/api/ReposApi.ts` — deleted 13 lines (`useRepositoryPRs` function + JSDoc + orphaned `CommitLog` import)
- `src/routes.tsx` — deleted 7 lines (`routePaths` reduce expression)

PR #360, opened 2026-04-16
refactor: remove dead useRepositoryPRs hook

wdeveloper16:
### Summary
- Removes `useRepositoryPRs` which is never imported anywhere in the codebase.

### Why it's safe to remove
- No imports of `useRepositoryPRs` exist outside its own definition.
- Also removes the now-orphaned `CommitLog` type import which was only needed by this hook.

### Changes
- `src/api/ReposApi.ts` — deleted 13 lines (the `useRepositoryPRs` function, its JSDoc comment, and the unused `CommitLog` import)

### DIFF
--- a/src/api/ReposApi.ts
+++ b/src/api/ReposApi.ts
@@ -1,7 +1,7 @@
 // Repository API hooks - uses /repos endpoints
 import { useApiQuery } from './ApiUtils';
 import { type RepositoryMaintainer, type RepositoryIssue } from './models';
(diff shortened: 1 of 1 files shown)


anderdc:
Closing — please combine this with #358 and #370 into a single dead-code cleanup PR. They're all the same type of change (removing unused hooks/exports) and don't need to be separate PRs. Keep #359 as the base since it already has the changes, and fold the others in.

PR #370, opened 2026-04-16
refactor: remove dead routePaths

wdeveloper16:
### Summary
- Removes `routePaths`, a route lookup map that is never imported anywhere in the codebase.

### Why it's safe to remove
- No imports of `routePaths` exist anywhere in the project, including tests.
- `getRouteForPathname` (retained, directly below) covers the same lookup pattern and is actively used.
- `routesArray` itself is retained — it is still used by the router and `getRouteForPathname`.

### Changes
- `src/routes.tsx` — deleted 7 lines (the `routePaths` reduce expression)

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -98,13 +98,6 @@ const routesArray: AppRoute[] = [
   },
 ];
 
-export const routePaths = routesArray.reduce<Record<string, AppRoute>>(
-  (acc, x) => {
-    acc[x.path] = x;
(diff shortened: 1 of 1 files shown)


anderdc:
Closing — please combine this with #358 and #360 into a single dead-code cleanup PR. They're all the same type of change (removing unused hooks/exports) and don't need to be separate PRs. Keep #359 as the base since it already has the changes, and fold the others in.

PR #387, opened 2026-04-16
Fix: Fix memory leaks

corevibe555:
Fix memory leaks.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] `npm run build` passes


anderdc:
**Closing -- these aren't memory leaks.**

A `setTimeout` that calls `setState` after unmount is not a memory leak. The timeout fires, the state setter is a no-op on an unmounted component, and the garbage collector cleans everything up normally. This was a React dev-mode console warning ("Can't perform a React state update on an unmounted component"), not an actual leak -- and React 18+ removed that warning entirely because the React team acknowledged it was misleading.

Adding 40 lines of ref tracking and cleanup effects across 4 files to prevent a no-op setState call has no user-facing benefit.

PR #435, opened 2026-04-16
fix: compute PR lines changed from per-PR data instead of unpopulated aggregate fields

YB0y:
## Summary

Fix the PRs stat card on the miner detail page showing "0 lines" for every miner. The card read `minerStats.totalAdditions` and `minerStats.totalDeletions`, but the `/miners/{githubId}` API endpoint never populates these fields (both return `null`). The per-PR data (`additions`/`deletions`) is already fetched by `useMinerPRs` in the same component.

The fix adds a `useMemo` that computes total lines from the already-loaded `prs` array and uses that value in the stat tile.

## Related Issues

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

**Before:** Screenshot 2026-04-16 141724

**After:** Screenshot 2026-04-16 141623

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/miners/MinerScoreCard.tsx
+++ b/src/components/miners/MinerScoreCard.tsx
@@ -324,6 +324,14 @@ const MinerScoreCard: React.FC<MinerScoreCardProps> = ({
     }, 0);
   }, [prs]);
 
+  const totalLinesChanged = useMemo(() => {
+    if (!prs || prs.length === 0) return 0;
+    return prs.reduce(
+      (sum, pr) => sum + (pr.additions || 0) + (pr.deletions || 0),
+      0,
+    );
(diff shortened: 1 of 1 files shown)


anderdc:
**Closing -- this should be fixed in the backend, not worked around on the frontend.**

The root cause is that the API doesn't populate `totalAdditions`/`totalDeletions` in the miner stats response. Summing per-PR data on the frontend is a band-aid that adds a redundant computation the backend should own. We'll fix this in das-gittensor or gittensor so the aggregate fields are populated correctly.

PR #417, opened 2026-04-16
feat(bounties): add sortable columns and align sorting UX with repository tables

carlh7777:
## Summary

- Added sortable column behavior to the bounties table in `IssuesList` (`ID`, `Repository`, `Issue`, and tab-specific columns like `Bounty`, `Status`, `Funding`, `Solver`, `Date`).
- Aligned sorting UX with repository tables: single active sort indicator, asc/desc toggle on repeated click, and full-header-cell hover/click interaction.
- Improved sorting performance by memoizing sorted output and precomputing sortable values per row (decorate/sort/undecorate), plus minor render optimization for repeated USD formatting.

### Layout improvements (from feedback / #421)
- Fixes stats layout responsiveness across breakpoints
- Improves mobile layout consistency based on feedback from #421

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

-before
image

-after
image
image

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #444, opened 2026-04-16
feat(leaderboard): add list view to miners leaderboard

### LINKED ISSUE #443: [Feature] Add list view to miners leaderboard
## Summary

Add a sortable table view to the Miner Leaderboard and Discoveries pages as an alternative to the card grid. A toggle in the page header switches between card and list layouts, and the selected view persists across navigation.

## Motivation

The card grid is low-density and hard to scan when comparing many miners. A table view — matching the existing Top Repositories table — gives users sortable columns, faster pivots on metrics like earnings or credibility, and a consistent UX across leaderboard pages.

bittoby:
## Summary

Adds a sortable list view to the Miner Leaderboard and Discoveries pages alongside the existing card grid, and extracts a **shared `DataTable` base component** (+ `useDataTableParams` hook) that `MinersList` is the first consumer of — the foundation for onboarding the remaining bespoke tables in a follow-up PR.

**List view UX (the feature):**

A small icon toggle in the page header switches between card grid and list view. The chosen view is remembered across navigation via `localStorage` (URL params still win when present, so shared links work). The list matches the existing Top Repositories table — same sticky header, cell styles, row hover. Every numeric column is sortable: clicking the active column flips asc/desc, clicking a different column resets to desc. Sort state (`sort`, `dir`) is URL-backed so specific views can be shared.

**Shared `DataTable` + `useDataTableParams` (the refactor):**

- `src/components/common/DataTable.tsx` — generic column-driven table with loading / error / empty states, optional row link, optional controlled sort, bundled MUI pagination config, and `header` / `footer` slots for richer toolbars.
- `src/hooks/useDataTableParams.ts` — URL-param state hook for sort / direction / page / rowsPerPage with per-field default direction overrides and configurable param keys (for pages hosting multiple tables). Pure helpers exported for unit testing without React DOM.
- `MinersList` is rebuilt as a declarative `DataTable` config — first real consumer, validating the API.

## Related Issues

Closes #443

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

[test.webm](https://github.com/user-attachments/assets/9594b592-bff3-48e0-910a-99326eb26d1e)

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/common/DataTable.tsx
+++ b/src/components/common/DataTable.tsx
@@ -1,0 +1,284 @@
+import React from 'react';
+import {
+  Alert,
+  Box,
+  CircularProgress,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+  TableSortLabel,
+  Typography,
+  type TableCellProps,
+} from '@mui/material';
+import { type SxProps, type Theme } from '@mui/material/styles';
+import { type SystemStyleObject } from '@mui/system';
+import { LinkTableRow } from './linkBehavior';
+import { bodyCellStyle, headerCellStyle, scrollbarSx } from '../../theme';
+import { type SortOrder } from '../../utils/ExplorerUtils';
+
+export type DataTableColumn<T, SortKey extends string = never> = {
+  /**
+   * Stable identifier for the column. Must be unique within the `columns`
+   * array (used as the React key).
+   */
+  key: string;
+  header: React.ReactNode;
+  width?: string | number;
+  align?: TableCellProps['align'];
+  renderCell: (item: T) => React.ReactNode;
+  headerSx?: SystemStyleObject<Theme>;
+  cellSx?: SystemStyleObject<Theme> | ((item: T) => SystemStyleObject<Theme>);
+  /**
+   * Setting this makes the column sortable. The actual "first-click" order
+   * for this key lives in the hook (`useDataTableParams.defaultOrderOverrides`);
+   * DataTable only renders the hover arrow direction.
+   */
+  sortKey?: SortKey;
+};
+
+export type DataTableSort<SortKey extends string> = {
+  field: SortKey;
+  order: SortOrder;
+  onChange: (nextField: SortKey) => void;
+};
+
+export type DataTableProps<T, SortKey extends string = never> = {
+  columns: DataTableColumn<T, SortKey>[];
+  rows: T[];
+  getRowKey: (item: T) => React.Key;
+  isLoading?: boolean;
+  isError?: boolean;
+  errorLabel?: string;
+  emptyLabel?: string;
+  emptyState?: React.ReactNode;
+  minWidth?: number | string;
+  /**
+   * When set, each row renders as an `<a href>` (native new-tab / middle-click
+   * support). NOTE: because the row becomes `<a>`, cells must not contain
+   * nested `<a>` elements (invalid HTML). For tables with nested anchors,
+   * use `onRowClick` instead.
+   */
+  getRowHref?: (item: T) => string;
+  linkState?: Record<string, unknown>;
+  /**
+   * Alternative to `getRowHref` — for tables that trigger dialogs, toggle
+   * expansion, or render nested `<a>` elements in cells. Ignored when
+   * `getRowHref` is provided.
+   */
+  onRowClick?: (item: T) => void;
+  getRowSx?: (item: T) => SystemStyleObject<Theme>;
+  header?: React.ReactNode;
+  /**
+   * Rendered below the table body when rows are present. Use this slot
+   * for pagination components (the app uses several styles — pluggable).
+   */
+  pagination?: React.ReactNode;
+  /**
+   * Rendered after the table regardless of loading / empty / error state.
+   * For summary bars or any controls that should persist across states.
+   */
+  footer?: React.ReactNode;
+  sort?: DataTableSort<SortKey>;
+  /**
+   * Defaults to false. Sticky header only resolves against a scroll
+   * ancestor with a bounded height — the consumer must provide one.
+   */
+  stickyHeader?: boolean;
+};
+
+const containerSx: SxProps<Theme> = {
+  overflowX: 'auto',
+  ...scrollbarSx,
+};
+
+const tableSx = {
+  tableLayout: 'fixed',
+  width: '100%',
+} as const;
+
+const clickableRowSx: SxProps<Theme> = (theme) => ({
+  cursor: 'pointer',
+  transition: 'background-color 0.2s',
+  '&:hover': {
+    backgroundColor: theme.palette.surface.subtle,
+  },
+});
+
+export const DataTable = <T, SortKey extends string = never>({
+  columns,
+  rows,
+  getRowKey,
+  isLoading = false,
+  isError = false,
+  errorLabel = 'Something went wrong loading this table.',
+  emptyLabel = 'No results to display.',
+  emptyState,
+  minWidth,
+  getRowHref,
+  linkState,
+  onRowClick,
+  getRowSx,
+  header,
+  pagination,
+  footer,
+  sort,
+  stickyHeader = false,
+}: DataTableProps<T, SortKey>) => {
+  const showTable = !isLoading && !isError && rows.length > 0;
+  const showEmpty = !isLoading && !isError && rows.length === 0;
+
+  return (
+    <>
+      {header}
+      {isLoading ? (
+        <Box sx={{ display: 'flex', justifyContent: 'center', py: 8 }}>
+          <CircularProgress />
+        </Box>
+      ) : null}
+      {isError && !isLoading ? (
+        <Alert severity="error" sx={{ m: 2 }}>
+          {errorLabel}
+        </Alert>
+      ) : null}
+      {showEmpty
+        ? (emptyState ?? (
+            <Box sx={{ p: 3 }}>
+              <Typography color="text.secondary">{emptyLabel}</Typography>
+            </Box>
+          ))
+        : null}
+      {showTable ? (
+        <>
+          <TableContainer sx={containerSx}>
+            <Table
+              stickyHeader={stickyHeader}
+              size="small"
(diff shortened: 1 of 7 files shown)


anderdc (changes requested):
The list view feature is a good idea, but the implementation adds another sortable header pattern when the repo already has two:

- **`SortableHeaderCell.tsx`** in `src/components/miners/` -- uses MUI's `TableSortLabel`, already shared
- **`TableSortLabel`** approach in `IssuesList.tsx` (#417)

Your `SortableHeader` in `MinersList.tsx` is a third hand-rolled implementation with manual arrow characters instead of MUI's built-in sort indicators.

Please reuse `SortableHeaderCell` (or generalize it if the types are too narrow) instead of adding another sortable header pattern. The repo already has 12+ table components -- we need to consolidate, not diverge further.

Crystora:
@anderdc I've updated. Pls review again.

Crystora:
@anderdc Pls give me any updates for me.

Crystora:
update

anderdc:
is there any way we can leverage other tables into this? I'm getting tired of all the unique table components we have. Maybe this PR can encapsulate creating some sort of shared component for all the future Tables and such. then another PR to flip the other tables to the new base component created with this PR.

Good PR, just need to start thinking about component composition more. Please start thinking on that before I add yet another table component.

Crystora:
Agreed, too many one-off tables. Quick audit, here's the plan:

**PR 1 - Base**
Extract `DataTable` + `useDataTableParams` hook. Migrate `SearchResultsTable` first (it's already column-driven, so lowest-risk validation).

**PR 2 - First real consumer**
Migrate `MinerPRsTable`. Has URL params + sort + pagination — real stress test for the API.

**PR 3 - This PR, reworked**
Rebase onto the new base, replace `MinersList` with a `DataTable` config. Ships the list-view feature without adding a bespoke component.

**PRs 4+ - Remaining migrations, one per PR**
`MinerRepositoriesTable`, `RepositoryPRsTable`, `RepositoryIssuesTable`, `IssueSubmissionsTable`, `IssuesList`, `TopMinersTable`.

**Skipping:**
- `RepositoryContributorsTable` - CSS grid, not a table
- `TopRepositoriesTable` - embedded chart needs splitting out first (separate epic)

Marking this PR as Draft and parking it until PR 1 + 2 land. I will open PR 1 soon next.

------------

Is that okay? Happy to your feedback

Crystora:
@anderdc Ready for review.

anderdc (changes requested):
**Good plan in the thread — `MinersList` consuming `DataTable` is exactly the right shape for a pilot. The DataTable API, however, has real gaps, and the hook isn't actually exercised. Please don't land this until it's piloting _both_.**

### Required before merge

- **Wire `useDataTableParams` into `TopMinersTable`.** Right now `MinersList` consumes `DataTable`, but `TopMinersTable` still hand-rolls every URL-param concern (`updateParam`, `handleSortChange`, `getSortOrderFromQuery`, `getViewModeFromQuery`, `getVisibleCountFromQuery`, the memoized reads for `sort`/`dir`). The hook is untouched by app code — only its pure helpers are tested. A real pilot replaces that logic with `useDataTableParams`. If the hook can't host `view=` / `eligible=` / `search=` / `visible=` alongside sort, that's the signal to expand the hook _now_, while MinersList is proving out the real constraints. Don't ship an API without a real consumer.

- **Drop the `<Card>` wrapper entirely. It's a table, not a card.** `DataTable` should render the `TableContainer` / `Table` / pagination / header+footer slots only — consumers wrap in `Card` (or `SectionCard`, or nothing) themselves. This kills `cardSx`, removes the need for any `disableCard` prop, and — critically — means each consumer chooses its own container height, which fixes `stickyHeader` too: sticky resolves against whatever scroll ancestor _they_ provide. Side benefit: `TopRepositoriesTable` (embedded chart) and any table composed inside another styled container can migrate without fighting the wrapper.

- **Flip `stickyHeader` default to `false`.** Once the Card is gone, sticky is opt-in per consumer who actually gives their container a bounded height. Today's default is silently broken for everyone.

- **Pick one source of truth for per-column default sort order.** `DataTableColumn.defaultSortOrder` and `useDataTableParams`'s `defaultOrderOverrides` encode the same concept and consumers have to keep them in sync. The comment `// Order applied the first time this column becomes active.` on `defaultSortOrder` is also misleading — DataTable only uses it to render the inactive-column hover arrow (line 193); the actual first-click order lives in the parent's `onChange`. Delete `DataTableColumn.defaultSortOrder`, keep `defaultOrderOverrides` in the hook, and have DataTable read from `sort` for the hover arrow too.

- **Pagination model doesn't match the app.** `MinerPRsTable`, `RepositoryPRsTable`, `RepositoryIssuesTable` use `src/components/miners/TablePagination.tsx` (minimal prev/next + `page/total`). DataTable hardcodes MUI's `TablePagination` with rows-per-page selector + first/last buttons. Migrating forces either a UX regression across every table or routing everyone through the `footer` slot and never using `pagination` (which also has the footer-visibility bug below). Resolve now — either swap DataTable's default to the existing `<TablePagination>`, or make it pluggable (`renderPagination?` prop).

- **Footer disappears on loading / empty / error; header doesn't.** `{header ? <Box>{header}</Box> : null}` renders unconditionally; `{footer}` lives inside the `showTable` branch. A ""load more"" or summary bar in `footer` gets asymmetric behavior. Render both always.

- **Style tokens fork silently.** `src/theme.ts` already exports `headerCellStyle` / `bodyCellStyle` (JetBrains Mono, uppercase, `surface.tooltip` bg). DataTable reinvents `headerCellSx` / `bodyCellSx` (no mono, no uppercase, `background.paper` bg). Post-migration every table visually changes. If intentional, state it and delete the now-dead exports in `theme.ts` and `leaderboard/types.ts`. Otherwise reuse the existing tokens.

- **No `onRowClick` escape hatch.** Only `getRowHref`. Tables that open dialogs, toggle expand, or trigger non-navigation actions can't use DataTable's row linking. Add `onRowClick?: (item: T) => void`.

- **Nested anchors will break on migration.** `MinerPRsTable` renders a nested `<a>` inside its row for the PR number with `stopPropagation`. Wrapping the row in `LinkTableRow` (`component=""a""`) produces invalid HTML `<a>…<a>…</a>…</a>`. Add a `linkBehavior: 'row' | 'none'` prop (or prefer `onRowClick` for that table), and document that rows rendered as `<a>` can't contain nested links.

### Design concerns

- **Zero render tests for `DataTable` and for the hook in a React tree.** Unit tests cover `parseSortField`, `computeNextSort`, etc., but the `useSearchParams`-integrated paths (memoized reads, `setSort` delete-vs-set branches, `paramKeys` override) and DataTable's render logic (sort click fires `onChange(sortKey)`, `emptyState` vs `emptyLabel` precedence, `getRowHref` → `<a>` rendering) are untested. For a foundation hosting 8+ tables, that's under-invested. A handful of `@testing-library/react` tests is cheap.

### Nits

- **A11y:** header `TableCell` has no `aria-sort`. `TableSortLabel` doesn't set it on the parent cell. Add `aria-sort={sort.field === sortKey ? (sort.order === 'asc' ? 'ascending' : 'descending') : 'none'}`.
- **Hover collision in `MinersList`:** `getRowSx` defines `&:hover: backgroundColor: border.subtle`; DataTable's `rowLinkSx` defines `surface.subtle`. MinersList wins (applied last). Pick one.
- **`MinersList.tsx:441`** — `transition: 'all 0.2s'` should be `background-color 0.2s`.
- **`column.key: string`** — no runtime uniqueness guarantee. Worth a TSDoc note.
- **`SortableHeaderCell` was generified but left in place.** Only used by `MinerRepositoriesTable`. Fine to leave until that table migrates, but please add a `// TODO: remove after MinerRepositoriesTable migrates to DataTable` so it doesn't calcify.

### Out of scope for this PR (noting for the migration roadmap)

Tables with per-row expand/collapse, inline editing, or column visibility toggles aren't covered by the current API. None of the eight listed tables need those today, so OK to defer — just flag if one turns out to need it.

Crystora:
@anderdc 
Updated PR based on your feedback:
- Wired `useDataTableParams` into `TopMinersTable` so all URL state (sort, visible, search, eligible, view) flows through the hook - not just sort.
- Dropped the `<Card>` wrapper from `DataTable` and flipped `stickyHeader` default to false - consumers now wrap themselves and own their scroll ancestor.
- Removed `DataTableColumn.defaultSortOrder` - `defaultOrderOverrides` in the hook is the single source of truth.
- Made pagination pluggable (`ReactNode` slot instead of hardcoded MUI).
- Footer always renders now, not just when the table body is visible.
- `DataTable` now reuses the canonical `headerCellStyle` / `bodyCellStyle` from `theme.ts` instead of forking.
- Added `onRowClick` as an escape hatch for tables with nested anchors, with keyboard a11y.
- Hit all the nits: `aria-sort` on header cells, narrowed `transition`, TSDoc on `column.key`, TODO on `SortableHeaderCell`, and dropped the hover collision in `MinersList`.

Pls review again. thanks

anderdc:
All prior asks addressed — thanks for the thorough rework. One last thing before I approve: this repo doesn't carry tests. Please remove `src/tests/useDataTableParams.test.ts` and any vitest / @testing-library dev-dep additions that exist solely to support it. Once that's gone I'll stamp it.

Crystora:
@anderdc Updated all! pls review once more

anderdc (changes requested):
Following up on the comment — this repo doesn't carry tests. Please remove `src/tests/useDataTableParams.test.ts` (and any vitest / @testing-library dev-dep additions that exist solely to support it). Everything else looks good; this is the last blocker.

Crystora:
> Following up on the comment — this repo doesn't carry tests. Please remove `src/tests/useDataTableParams.test.ts` (and any vitest / @testing-library dev-dep additions that exist solely to support it). Everything else looks good; this is the last blocker.

Hi, I've already removed vitest/@testing-library. Pls review carefully again.

anderdc (approved):

PR #453, opened 2026-04-17
fix: guard against non-array API responses crashing .map() calls

### LINKED ISSUE #454: Fix guard against non-array API responses crashing .map() calls
## Description

## Steps to Reproduce

1. Navigate to `/bounties`

## Expected Behavior

## Actual Behavior

Affected locations:

| File | Line | Error |
|---|---|---|

## Environment

- Browser: Chrome
- OS: Linux

## Additional Context

```ts
data || []
data ?? []

```

Image

dataCenter430:
## Summary

Replaces `data || []` and `data ?? []` fallbacks with `Array.isArray()` guards in three locations. These patterns silently pass a truthy non-array object through unchanged, causing `TypeError: .map is not a function` and triggering the error boundary.

**Affected files:**
- `src/pages/IssuesPage.tsx` — 3 `data || []` guards on issues query results
- `src/pages/dashboard/useDashboardData.ts` — 3 `data ?? []` guards for prs/miners/issues
- `src/api/SearchApi.ts` — chained `datasetQuery.data ?? cachedData ?? []` guard

## Test plan
- [ ] Navigate to `/bounties` — issues list renders normally
- [ ] Open global search bar while miners data is loading — no crash
- [ ] Dashboard renders with all three datasets loading

## Screenshots (After fix)

image

### DIFF
--- a/src/api/SearchApi.ts
+++ b/src/api/SearchApi.ts
@@ -44,7 +44,11 @@ const useCachedSearchDataset = <T>(
   );
 
   return {
-    data: datasetQuery.data ?? cachedData ?? [],
+    data: Array.isArray(datasetQuery.data)
+      ? datasetQuery.data
+      : Array.isArray(cachedData)
+        ? cachedData
+        : [],
     isLoading: isDatasetLoading(
       shouldFetch,
       cachedData,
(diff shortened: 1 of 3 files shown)


anderdc:
Closing — this is overprotective coding. The API client and its callers already guarantee these responses are arrays; wrapping every `.map()` site with `Array.isArray` guards is defensive padding for a scenario that can't happen. If a response were ever genuinely non-array, the fix belongs at the API boundary, not at every consumer.

PR #333, opened 2026-04-15
refactor: extract shared PR table helpers and remove duplicate status filter type

statxc:
## Summary

`MinerPRsTable` and `RepositoryPRsTable` each had their own inline status filtering chains and pagination slicing. This PR extracts the duplicated logic into `src/utils/prTable.ts` (`filterPrs()`, `paginateItems()`, `PrStatusFilter`) so both tables consume a single shared implementation.

Also removes the duplicate `MinerStatusFilter` type from `ExplorerUtils` — it was identical to `PrStatusFilter`.

No visual or behavioral changes.

## Related Issues

N/A

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

> No UI changes — pure refactor.

PR #456, opened 2026-04-17
fix: add Array.isArray guards to prevent .map() crashes on non-array API responses

### LINKED ISSUE #454: Fix guard against non-array API responses crashing .map() calls

tarai-dl:
Closes #454

## Problem

Multiple pages crash with `TypeError: <variable>.map is not a function` when the API returns a non-array response (e.g., `{ items: [...], total: 100 }`). The `data || []` and `data ?? []` patterns don't protect against truthy non-array values.

## Fix

- `IssuesList.tsx`: guard `rawIssues` prop with `Array.isArray` before `.map()`
- `IssuesPage.tsx`: replace `data || []` with `Array.isArray(data) ? data : []`
- `searchData.ts`: guard `miners` parameter in `buildMinerSearchData`
- `useDashboardData.ts`: guard `prs`, `miners`, `issues` data with `Array.isArray`

All data sources are now safely normalized to arrays before use.


dataCenter430:
Hey, @statxc 
I've already raised the PR for #454 .
Your PR is duplicate with my PR https://github.com/entrius/gittensor-ui/pull/453

statxc:
Hey @dataCenter430, ping the right owner. I'm not concerned about this PR!

dataCenter430:
> Hey @dataCenter430, ping the right owner. I'm not concerned about this PR!

Please fix the commit history. 😄 
image

statxc:
Hey @dataCenter430, I’d strongly recommend reviewing all more carefully yourself before pinging someone.
My commit came from PR #333, which was a completely separate PR and was already merged into the test branch.
tarai-dl branched off `test` (which already had my commit) but opened their PR against `main` (which didn’t). That’s why my commit is showing up in their PR.

dataCenter430:
Ah, sorry for the mistake. @statxc 
And thanks for your effort.

anderdc:
Closing — #453 is the correct PR for this.

Reasons:
- **Duplicate of #453** (by @dataCenter430) which addresses the same issue (#454) with a tighter scope (3 files, 23/7) vs this one (9 files, 86/66).
- **Targets `main` instead of `test`.** All UI PRs must target `test`.
- The extra ~50 lines of diff on `MinerPRsTable.tsx` / `RepositoryPRsTable.tsx` / new `prTable.ts` aren't Array-guard changes — they're from #333 (extract shared PR table helpers), which is already merged on `test`. They only show up here because this PR branched off stale `main`.

Per our duplicate-PR policy we keep the correctly-scoped, correct-branch PR. Re-open against `test` only if #453 is rejected and we need an alternative.

PR #473, opened 2026-04-17
feat: add retry/timeout logic to ApiUtils

### LINKED ISSUE #477: Add retry logic and request timeout to useApiQuery
## Summary
`useApiQuery` in `src/api/ApiUtils.ts` currently has `retry: false` and no request timeout. Any failed request — rate limit (429), transient 5xx, or a
slow server response — fails immediately and permanently, leaving the UI in a blank or broken state with no recovery attempt.

## Motivation
All API calls in the app go through `useApiQuery`, so a single fix here improves resilience across every page and component. 
Transient infrastructure issues (network blips, server restarts, momentary overload) are currently surfaced to users as apparent product bugs because the UI never retries. 
A request timeout is also missing, meaning a hanging server response blocks the UI indefinitely with no fallback.

Retrying only on recoverable status codes (429, 5xx) with exponential backoff, and adding a request timeout, would silently recover the majority of these cases without any changes to individual callers.

dataCenter430:
## Summary

- Adds `REQUEST_TIMEOUT_MS = 15_000` (15 s) to all Axios requests via `useApiQuery`
- Adds `shouldRetry` — retries on 429, 5xx, and network errors; stops after `MAX_RETRIES = 3`; never retries 4xx client errors
- Adds `getRetryDelay` — exponential backoff capped at 30 s: 1 s → 2 s → 4 s → … → 30 s
- Wires `retry` and `retryDelay` into React Query options in `useApiQuery`
- Adds 20 unit tests in `src/tests/ApiUtils.test.ts` (vitest)

## Test plan

- [x] `npx vitest run` — 57/57 tests pass (no regressions)
- [x] TypeScript — zero type errors
- [x] ESLint — no warnings
- [x] Prettier — formatted correctly

### DIFF
--- a/src/api/ApiUtils.ts
+++ b/src/api/ApiUtils.ts
@@ -1,6 +1,25 @@
 import { useQuery } from '@tanstack/react-query';
 import axios, { type AxiosError } from 'axios';
 
+const REQUEST_TIMEOUT_MS = 15_000;
+export const MAX_RETRIES = 3;
+
+const isRetryableStatus = (status: number): boolean =>
+  status === 429 || status >= 500;
+
+export const shouldRetry = (
+  failureCount: number,
+  error: AxiosError,
+): boolean => {
+  if (failureCount >= MAX_RETRIES) return false;
+  const status = error?.response?.status;
+  if (status === undefined) return true; // network error or timeout — no HTTP response
+  return isRetryableStatus(status);
+};
+
+export const getRetryDelay = (attempt: number): number =>
+  Math.min(1_000 * 2 ** attempt, 30_000);
+
 export const useApiQuery = <TResponse = void, TSelect = TResponse>(
   queryName: string,
   url: string,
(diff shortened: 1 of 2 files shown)


anderdc:
Closing along with #477.

This is overly defensive — retry/timeout logic built against a theoretical failure mode nobody has actually hit. No incident, no user reports, no citation of transient 5xx or rate-limit issues. gittensor-ui talks to a validator-controlled backend; we're not seeing the kind of transient failures this guards against.

The tradeoffs also aren't great:
- `retry: 3` + 15 s timeout means a real outage spins for ~60 s before surfacing to the user — worse UX than fast-fail.
- Silent retries mask real backend problems from monitoring.
- Bespoke `shouldRetry` / `getRetryDelay` + 108 lines of tests is a lot of surface area for a hypothetical.

If a concrete incident ever drives this, a one-line `retry: 3` handles 99% of it.

dataCenter430:
Hi, @anderdc 
Fair points, especially on UX. A 15 s timeout × 3 retries does mean a real outage takes up to ~60 s to surface, which is worse than fast-fail for users.

I'd push back slightly on one thing: the timeout alone has no UX downside and is independently valuable. Without it, a hung connection never resolves — the spinner runs forever. A 15 s hard cut is strictly better than no timeout regardless of retry policy.

Would you consider a slimmed-down version that drops the retry entirely and just adds timeout: 15_000 to the Axios call? That's a one-liner change, no bespoke logic, no test surface area — and it closes the "request hangs forever" hole without the silent-retry tradeoffs.

Happy to reopen with only that change if it sounds reasonable.

Would be it possible to consider this again?

PR #312, opened 2026-04-15
fix: back button and scroll-to-top issues on tabbed pages

YB0y:
## Summary

- Fix back button requiring multiple clicks on Miner Details page by using `{ replace: true }` on all `setSearchParams` calls for tabs, filters, and pagination
- Fix scroll jumping to top when switching tabs by stabilising the `useOnNavigate` callback with `useRef` so it only fires on actual pathname changes
- Add `{ replace: true }` to OnboardPage tab navigation for consistency

## Related Issues

Fixes #311

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Videos
Bug:
https://github.com/user-attachments/assets/fe526026-1fa6-4a10-96b7-a041f0ada75c
Fix:
https://github.com/user-attachments/assets/c6292bc5-9015-4770-b3a8-64bfb525acd7

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [ ] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #201, opened 2026-04-14
fix: persist Overview tab score breakdown page number across navigation

leonaIee:
## Summary

Persist the Overview tab's Score Breakdown page number in URL search params (`scorePage`) so navigating to a PR detail page and back preserves the current page instead of resetting to page 1.

Same approach as PR #198 but for the Overview tab's `MinerScoreBreakdown` component.

## Related Issues

Fixes #197

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

https://github.com/user-attachments/assets/da272a6a-64a6-4960-b3bf-a974c752118c

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #486, opened 2026-04-17
fix(repositories): persist PR status filter in URL

### LINKED ISSUE #484: RepositoryPRsTable status filter is not persisted in URL
## Description

In Repository detail Pull Requests tab, the status filter (All, Open, Merged, Closed) is managed only in local React state and is not synced to URL query params.
As a result, the selected filter is not preserved on refresh and cannot be shared via link.

## Steps to Reproduce

1. Open a repository details page that renders RepositoryPRsTable.
2. In the PR table, click a status filter (for example, Open or Merged).
3. Refresh the page (or copy/open the same URL in a new tab).

## Expected Behavior

The selected filter should be reflected in the URL (for example ?prStatus=open) and remain selected after refresh/navigation.
Users should be able to share a URL that opens with the same filter applied.

## Actual Behavior

The table filters correctly in-session, but the URL does not change.
After refresh or reopening the page, filter resets to default (All).

## Environment

- Browser: Any
- OS: Any

leonaIee:
## Summary

In `RepositoryPRsTable` (rendered on the Repository Details → Pull Requests tab), the status filter (All / Open / Merged / Closed) was managed only in local React state. Refreshing the page or copying the URL dropped the selection and fell back to the default filter.

Fix syncs the selection to the `prStatus` URL search param on click and reads it back on mount — the same pattern already used by `MinerPRsTable` and previously approved by the maintainer in #198, #201, and #299.

- URL param name `prStatus` matches the existing convention in `MinerPRsTable`
- `isPrStatusFilter` type-guard rejects stale/invalid values so malformed URLs fall back to the component's `state` prop default
- `setSearchParams` uses `{ replace: true }` (approved pattern from #312)
- `prStatus=all` is omitted from the URL when filter is `'all'` to keep the URL clean

No behavior change on mount when the URL has no `prStatus` — `filter` still falls through to the `state` prop default, so any parent that passes `state="open"` or similar is unaffected.

## Related Issues

Fixes #484

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/repositories/RepositoryPRsTable.tsx
+++ b/src/components/repositories/RepositoryPRsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import {
   Card,
   Typography,
@@ -29,7 +29,7 @@ type PrSortField =
   | 'mergedAt';
 type SortOrder = 'asc' | 'desc';
 import { useAllPrs } from '../../api';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 import {
   TEXT_OPACITY,
   scrollbarSx,
@@ -38,6 +38,16 @@ import {
 } from '../../theme';
 import { filterPrs, getPrStatusCounts, type PrStatusFilter } from '../../utils';
 import FilterButton from '../FilterButton';
+
+const PR_STATUS_FILTERS: readonly PrStatusFilter[] = [
+  'all',
+  'open',
+  'merged',
+  'closed',
+];
+
+const isPrStatusFilter = (value: string | null): value is PrStatusFilter =>
+  value !== null && (PR_STATUS_FILTERS as readonly string[]).includes(value);
 
 interface RepositoryPRsTableProps {
   repositoryFullName: string;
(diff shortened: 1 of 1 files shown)


peter-rango:
Hi @anderdc — whenever your queue opens up, I'd really appreciate a look at this one.

This follows the exact URL-persistence pattern you approved in #198 (*"Clean approach, single source of truth from search params"*), #201 (*"Same clean pattern as #198"*), and #299. The diff is intentionally small (**+33 / −3 in one file**) and mirrors `MinerPRsTable`'s `prStatus` flow almost line-for-line:

- Same URL param name (`prStatus`) as the existing `MinerPRsTable` convention
- `isPrStatusFilter` type-guard so malformed URLs fall back to the component's `state` prop default — no regression for parents passing `state="open"`
- `setSearchParams` uses `{ replace: true }` (the pattern you endorsed in #312)
- `prStatus=all` is omitted from the URL to keep things tidy

No behavior change on mount when there's no URL param — it still falls through to the existing default. Only the new "survive refresh / shareable link" behavior is added. Closes #484.

Happy to iterate — for example, if you'd prefer `PR_STATUS_FILTERS` + `isPrStatusFilter` extracted into `src/utils/prStatus.ts` to dedupe with `MinerPRsTable`, I kept it local to avoid scope creep but it's a natural follow-up. Thanks for taking a look!

PR #529, opened 2026-04-17
feat(ui): remember last tab on bounties, miner, and PR detail pages

### LINKED ISSUE #508: [Feature] Save last tab position
On a PR that touches many files, finding a specific identifier (a constant name, a function, a flag) currently means scrolling the file tree and opening each candidate by hand. Most diff viewers ship some form of in-diff search out of the box.

Expected behavior:

- A search field above the file tree, with a clear button and an inline case-sensitive toggle.
- The tree filters to files that contain the query, with a per-file match count next to the name.
- Matching substrings are highlighted inside every rendered diff line, in both Split and Unified views.
- A short status line under the input shows total matches and matching file count.
- Empty state echoes the query when nothing matches.
- The query persists through the existing Split/Unified, Wrap Lines, and Only Changed toggles.

hook-ops:
## Summary

This PR makes multi-tab pages feel stateful instead of resetting on every visit. We remember the user’s last selected tab in `localStorage` (per browser profile) for **Issue Bounties**, **Miner details**, and **PR details**.

## Problem

Several important flows use tabs (bounties sub-pages, miner sections, PR sub-views). Previously, revisiting a page always snapped back to the default tab, even if the user consistently works from a different tab. That forces repeated manual clicks and breaks flow.

## Solution

We persist the most recently selected tab and restore it on the next visit **only when the URL does not already specify a valid tab**. If the URL includes an explicit tab (path segment or query param), that value always wins—this keeps deep links and shared URLs predictable.

### Issue Bounties (`IssuesPage`)

- Route is `/bounties/:tab?`. If the user hits `/bounties` with no tab, or uses an invalid slug, we redirect to the last saved tab (fallback `available`).
- If the URL already contains a valid tab (`available`, `pending`, `history`), we use it and refresh the saved preference.

### Miner details (`MinerDetailsPage`)

- Tabs are driven by the `tab` query string and differ between **OSS Contributions** vs **Issue Discovery** (different tab sets).
- We store the last tab **separately per mode** so switching modes does not clobber the other mode’s preference.
- Switching between OSS / Issue Discovery now restores that mode’s remembered tab instead of always forcing `overview`.

### PR details (`PRDetailsPage`)

- Adds `?tab=overview|files|conversation` so the active sub-view is representable in the URL.
- On first load for a PR (after details are available), if no valid `tab` is present, we restore the last saved tab and sync the URL.
- Tab clicks update both UI state and `localStorage`.

### Shared utility

- `src/utils/tabPreferences.ts` centralizes safe read/write to `localStorage` (handles private mode / quota failures gracefully).

## Related issue

Fixes #508

## Follow-ups (optional)

Repository details and other tabbed surfaces can reuse the same approach in a separate PR if maintainers want full coverage.

## Test plan

- Bounties: visit `/bounties`, choose **Pending**, reload → still **Pending**; open `/bounties/history` directly → **History** wins; return to `/bounties` → last saved tab.
- Miner: pick a non-default tab, reload → preserved; open with explicit `?tab=` → URL wins; toggle OSS vs Issue Discovery → each mode remembers independently.
- PR: pick **Files**, reload → preserved; open with `?tab=conversation` → **Conversation** wins.

## Commands run locally

npm run build
npm test


anderdc:
Visible UX changes (remember-last-tab behavior is user-observable) must include a screenshot or short video per the contribution guidelines. Also currently has merge conflicts. Please re-open with media and a clean rebase.

PR #541, opened 2026-04-17
refactor: keep consistency for tables across all pages

### LINKED ISSUE #539: fix: unify table header and body cell styles across bounties and languages pages
## Description

The table style should be aligned across all tables in gittensor platform.

## Steps to Reproduce

1. Go to repositories page and observe that the repositories table style including hearderCell.
2. Go to onboard page and observe that languages table style including headerCell.
3. Go to bounties page and observe that the tables style for Available issues and History tab

## Expected Behavior

The table styles are different for three tables.

## Actual Behavior

The table style should be aligned across all tables in gittensor platform.

## Environment

- Browser: Chrome
- OS: Linux

## Additional Context

https://github.com/user-attachments/assets/a4ec898a-52bc-4338-9a87-423e133979b6

### LINKED ISSUE #540: Add header section including chart toggle, Rows selection and search input for bounties tables.
## Summary

Add header section including chart toggle, Rows selection and search input for bounties tables.

## Motivation

I think the tables should be aligned for the styles across all pages in the gittensor platform.
To keep the same design for the tables, we need to follow the repositories table style for bounties table.
Remove Available Issues, Pending Issues and History tabs and add header section including chart toggle, Rows selection and search input like repositories page.

Image

Image

dataCenter430:
## Summary

- **Unified table styles** across all pages (bounties, languages) using shared `headerCellStyle` / `bodyCellStyle` from `theme.ts`, matching the existing repositories table design
- **Bounties page redesign**: removed the Available / Pending / History tab navigation and replaced with a single unified table featuring an inline toolbar (filter buttons, chart toggle, rows-per-page selector, search input) — consistent with the repositories page pattern
- **Fixed column width instability** on the bounties table: switching rows-per-page (10→25→50) no longer causes the Repository column to expand and push the Date column off-screen; resolved by enforcing `table-layout: fixed` with explicit column widths and text truncation on the Repository cell
- **Global search bar polish**: made the sticky search bar container full-width, centered, with symmetric vertical padding and a bottom border to cleanly separate it from page content below

## Files changed

| File | Change |
|---|---|
| `src/components/issues/IssuesList.tsx` | Full rewrite — unified styles, inline filter toolbar, chart toggle, fixed table layout |
| `src/pages/IssuesPage.tsx` | Removed tab state, combined three API queries into a single flat array |
| `src/routes.tsx` | Simplified bounties route from `/bounties/:tab?` to `/bounties` |
| `src/components/repositories/LanguageWeightsTable.tsx` | Applied `headerCellStyle` / `bodyCellStyle` from theme; removed local inline style objects |
| `src/components/layout/AppLayout.tsx` | Global search bar container: full-width, centered, symmetric padding, bottom border |

## Test plan

- [ ] Bounties page loads with all issues visible under "All" filter
- [ ] Filter buttons (Available / Pending / History) correctly subset the table
- [ ] Switching rows-per-page (10 → 25 → 50) keeps all columns visible including Date
- [ ] Chart toggle shows/hides the bar chart with animation
- [ ] Search input filters by repository name and issue title
- [ ] Table header and body cells match the repositories page visual style
- [ ] Languages page table uses the same header/body cell styles
- [ ] Global search bar is centered with a visible bottom border on all pages that show it
- [ ] Sticky search bar does not overlap table content when scrolling
- [ ] `/about` and `/faq` redirect routes still work

## Screenshot

Before:

https://github.com/user-attachments/assets/78845c48-d1a1-4a66-9fa5-167a141e7776

After:

https://github.com/user-attachments/assets/a75f59ee-3073-4897-85ad-8a36143919e2

### DIFF
--- a/src/components/issues/IssuesList.tsx
+++ b/src/components/issues/IssuesList.tsx
@@ -622,17 +622,8 @@ const IssuesList: React.FC<IssuesListProps> = ({
     );
   }
 
-  return (
-    <Card
-      sx={{
-        backgroundColor: 'background.default',
-        border: `1px solid ${theme.palette.border.light}`,
-        borderRadius: 3,
-        overflow: 'hidden',
-      }}
-      elevation={0}
-    >
-      {/* Header toolbar */}
+  const headerToolbar = (
+    <>
       <Box
         sx={{
           px: 2,
@@ -785,7 +776,6 @@ const IssuesList: React.FC<IssuesListProps> = ({
         </Box>
       </Box>
 
-      {/* Chart */}
       <Collapse in={showChart}>
         <Box
           sx={{
@@ -803,40 +793,47 @@ const IssuesList: React.FC<IssuesListProps> = ({
           )}
         </Box>
       </Collapse>
+    </>
+  );
 
-      {filteredIssues.length === 0 ? (
-        <Box sx={{ p: 4, textAlign: 'center' }}>
-          <Typography
-            sx={{
-              color: alpha(theme.palette.common.white, TEXT_OPACITY.tertiary),
-            }}
-          >
-            {searchQuery ? 'No issues match your search' : 'No issues found'}
-          </Typography>
-        </Box>
-      ) : (
-        <>
-          <DataTable<IssueBounty, SortKey>
-            columns={columns}
-            rows={paginatedIssues}
-            getRowKey={(issue) => issue.id}
-            getRowHref={
-              getIssueHref ? (issue) => getIssueHref(issue.id) : undefined
-            }
-            linkState={linkState}
-            minWidth={
-              filterType === 'history'
-                ? '1000px'
-                : filterType === 'pending'
-                  ? '900px'
-                  : '750px'
-            }
-            sort={{
-              field: sortKey,
-              order: sortDirection,
-              onChange: handleSort,
-            }}
-          />
+  return (
+    <Card
+      sx={{
+        backgroundColor: 'background.default',
+        border: `1px solid ${theme.palette.border.light}`,
+        borderRadius: 3,
+        overflow: 'hidden',
+      }}
+      elevation={0}
+    >
+      <DataTable<IssueBounty, SortKey>
+        columns={columns}
+        rows={paginatedIssues}
+        getRowKey={(issue) => issue.id}
+        getRowHref={
+          getIssueHref ? (issue) => getIssueHref(issue.id) : undefined
+        }
+        linkState={linkState}
+        minWidth={
+          filterType === 'history'
+            ? '1000px'
+            : filterType === 'pending'
+              ? '900px'
+              : '750px'
+        }
+        header={headerToolbar}
+        emptyState={
+          <Box sx={{ p: 4, textAlign: 'center' }}>
+            <Typography
+              sx={{
+                color: alpha(theme.palette.common.white, TEXT_OPACITY.tertiary),
+              }}
+            >
+              {searchQuery ? 'No issues match your search' : 'No issues found'}
+            </Typography>
+          </Box>
+        }
+        pagination={
           <TablePagination
             rowsPerPageOptions={[]}
             component="div"
@@ -848,8 +845,13 @@ const IssuesList: React.FC<IssuesListProps> = ({
             showFirstButton
             showLastButton
           />
-        </>
-      )}
+        }
+        sort={{
+          field: sortKey,
+          order: sortDirection,
+          onChange: handleSort,
+        }}
+      />
     </Card>
   );
 };
--- a/src/components/issues/IssuesList.tsx
+++ b/src/components/issues/IssuesList.tsx
@@ -1,17 +1,31 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
 import {
   Avatar,
   Box,
   Card,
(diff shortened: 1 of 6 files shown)


dataCenter430:
Hi, @anderdc 
Would be it possible to review the PR?
Note: Please take a look at the screenshots to help the understand.

dataCenter430:
Hi, @anderdc 
Would be it possible to review the PR asap?

anderdc (changes requested):
Direction looks good — shared `headerCellStyle` / `bodyCellStyle`, `table-layout: fixed`, and the unified toolbar are all clean wins. Three must-fixes before merge:

1. **Deep-linking broken.** `/bounties/:tab?` → `/bounties` in `src/routes.tsx`. Existing bookmarks and external links to `/bounties/history`, `/bounties/pending`, `/bounties/available` now 404. Please promote filter to a query param (`?filter=history`) and add route redirects for the three legacy paths so old links keep working.

2. **No dedup on merged arrays** in `src/pages/IssuesPage.tsx`. `[...activeIssuesQuery.data, ...registeredIssuesQuery.data, ...historyIssuesQuery.data]` trusts perfect partitioning from three separate queries. Any transient overlap between polls will render the same `issue.id` twice and corrupt sorting/pagination. Dedup by id before passing to `IssuesList`.

3. **Loading UX regression.** `isLoading = activeIssuesQuery.isLoading || registeredIssuesQuery.isLoading || historyIssuesQuery.isLoading` blocks the whole page until the slowest query resolves. Previously each tab loaded independently. Prefer rendering what's available as each query resolves.

dataCenter430:
Hi, @anderdc 
Thanks for your review.
I've fixed the issues. Ready to review.

anderdc (changes requested):
Thanks for the quick turnaround and the clean rebase onto #545. Two of the three blockers are resolved; one is half-resolved.

**Fix 2 (dedup) — ✅ resolved.** `Set`-based dedup by `issue.id` in `IssuesPage.tsx` is correct.

**Fix 3 (loading UX) — ✅ resolved.** Flipping `||` → `&&` so the skeleton only shows while *no* query has data and the first to resolve unblocks partial rendering is exactly right.

**Fix 1 (deep-linking) — still broken.** The filter-from-URL logic in `IssuesList.tsx` (via `useSearchParams`) is correct, but the legacy-path redirect can't actually run:

- `routes.tsx:48` currently has `path: '/bounties'`.
- `IssuesPage.tsx` adds a `useEffect` that reads `useParams<{ tab?: string }>()` and redirects `/bounties/history` → `/bounties?filter=history`.
- In React Router v6, route `/bounties` does **not** match `/bounties/history` / `/bounties/pending` / `/bounties/available` — `IssuesPage` never mounts on those URLs, so the effect never runs and `tabParam` is always `undefined`. Legacy bookmarks still 404.

**Smallest correct fix:** restore `path: '/bounties/:tab?'` in `routes.tsx`. `IssuesPage` will then mount on legacy paths, `tabParam` will be populated, and the existing redirect effect does its job. The rest of the fix (query-param filter, `useSearchParams` as source of truth) is already in place and doesn't need to change.

One more pass and this is good to go.

dataCenter430:
Just done!
Now it's clean and ready to review.
Thanks, @anderdc

anderdc (changes requested):
Route fix from `70467238` looks good — legacy deep-links now work end-to-end. One more round before approval, because this PR's stated purpose is consistency across tables and the current shape doesn't leverage the shared pieces #545 just landed.

**Ask: move toolbar, pagination, and empty state into `DataTable`'s slots.**

Every other migrated table in #545 uses this pattern — see `MinerPRsTable.tsx`, `RepositoryPRsTable.tsx`, and `SearchResultsCard.tsx`. IssuesList should match:

- **Toolbar** (filter buttons + chart toggle + rows select + search input) — move into `<DataTable header={headerToolbar}>` instead of rendering as a sibling `<Box>` inside the `<Card>`.
- **Chart `<Collapse>`** — can live inside the `header` block, or immediately above DataTable; just keep it off the raw Card body so DataTable owns the full table surface.
- **Pagination** — move `<TablePagination>` into `<DataTable pagination={<TablePagination …/>}>` rather than rendering it as a sibling below.
- **Empty state** — replace the `filteredIssues.length === 0 ? <Box>…</Box> : <DataTable …/>` ternary with DataTable's `emptyLabel` (or `emptyState` if you need the searchQuery-aware copy variant).

Net effect: ~50 LOC reduction in `IssuesList.tsx` and visual/behavioral consistency with every other table in the app. This matters because #541's whole premise is "keep consistency for tables across all pages" — falling short of the shared-component pattern here undercuts the PR's own stated goal.

**Broader principle:** lean on existing components wherever they exist. If there's a shared piece (`FilterButton`, `DataTable`, theme tokens, `LinkBox`), prefer composing it over reinventing the layout locally. Rule of thumb: if two other tables in the app already solve this exact shape, use that shape.

After this, approve.

anderdc:
Also, I think I'd rather the header style in the bounties table to be used across all the tables, if you can change to use that instead along with final review

dataCenter430:
Good catch, 👍 
Okay, I will fix the PR as you wanted.
Thanks

dataCenter430:
Hi, @anderdc

Just done, All fixed.
Ready to review again.
Thanks.

anderdc (approved):

PR #591, opened 2026-04-18
fix(pagination): make prev/next controls keyboard accessible

fansilas:
## Summary

The Prev and Next controls in the shared `TablePagination` were plain `<Box>` elements with only an `onClick`, so keyboard focus skipped past them and screen readers had no role or label to announce. Since this component backs the miners, repositories, and PR tables, pagination was effectively unreachable without a mouse on most list views.

Swapping the two Boxes for MUI `<IconButton>` fixes it with the native primitive: focus order, Enter/Space activation, `aria-label` ("Previous page" / "Next page"), and a real `disabled` state that also drives `aria-disabled`. The existing visual treatment (0.3 opacity at the boundary pages, hover color) is preserved via `&.Mui-disabled` and the hover rule.

## Related Issues

Fixes #590 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

No visual change. Before and after look identical; the difference is in the a11y tree and tab order.

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

### Manual verification

1. Opened the miners leaderboard with enough rows to paginate.
2. Tabbed through the table footer: focus now lands on Prev, then Next, with a visible focus ring.
3. Pressed Enter and Space on each control; both advance the page.
4. On the first page, Prev is programmatically disabled (not just dimmed); same for Next on the last page.
5. VoiceOver announces "Previous page, button" and "Next page, button", with the dimmed state read as disabled.

PR #593, opened 2026-04-18
fix(a11y): make copy controls keyboard reachable and clipboard safe

### LINKED ISSUE #592: [BUG] Copy buttons in Getting Started and PR Files Changed claim success even when the clipboard write fails
## Description

Several copy controls in the app write to the clipboard without checking the result and without any accessibility affordances, so they report success on every click even when nothing was copied.

1. **Getting Started code block copy icon** is a `<Box onClick=…>`. It has no `role`, no `tabIndex`, no `aria-label`, and no key handler. The handler calls `navigator.clipboard.writeText(children.trim())` with no try/catch and no await, then flips `copied` to `true` unconditionally.

2. **PR Files Changed copy-path icon** is a proper `IconButton`, but the handler also calls `navigator.clipboard.writeText(file.filename)` without awaiting or catching rejections, and `setCopied(true)` fires regardless of outcome. The `IconButton` has no `aria-label` of its own, so assistive tech relies solely on the Tooltip title, which is not exposed until the tooltip is visible.

Both behave incorrectly on any host where the Clipboard API is unavailable or rejected (non-secure contexts, restricted iframes, denied permission): the check-mark appears and the tooltip reads "Copied!" even though the clipboard was never updated. The Getting Started button is additionally unreachable by keyboard.

For reference, `MinerScoreCard` already handles the same flow correctly (`ButtonBase`, `aria-label`, `aria-live`, try/catch around `navigator.clipboard.writeText`, cleanup timer). Aligning the other two sites with that pattern would fix the behaviour and make the UX consistent across the app.

## Steps to Reproduce

1. Open the Onboard -> Getting Started page on any step that shows a command block.
2. With the mouse put aside, try to reach the copy icon with Tab.
3. Open any PR details page, expand "Files Changed", and hover a file header.
4. Open either page over plain `http://`, or inside a restricted iframe, or with clipboard permission denied in DevTools.
5. Click the copy icon on both pages.
6. Observe the tooltip and icon state after each click.
7. Separately, walk through both headers with a screen reader (VoiceOver, NVDA) and listen to how the copy buttons are announced.

## Expected Behavior

- Both copy controls are reachable by keyboard and announce a clear name such as "Copy command, button" or "Copy file path, button".
- Enter or Space activates the focused control.
- On successful clipboard write, the check-mark appears and the tooltip reads "Copied!".
- On clipboard failure or rejection, the UI does not pretend the copy succeeded. Reverting to the original icon or surfacing a subtle error both work.
- The source text stays selectable so manual copy remains possible as a fallback.

## Actual Behavior

- Getting Started: Tab skips the copy icon entirely, no accessible name, and every click flips the UI to the success state regardless of outcome.
- PR Files Changed: keyboard focus works, but the button lacks its own `aria-label`, and every click still flips to the success state even when the clipboard rejected the write.
- Clipboard rejections are swallowed silently in both components.

## Environment

- Browser: any (reproduces most easily on non-secure contexts or when clipboard permission is denied)
- OS: any

## Additional

The fix is a good fit for a single shared primitive. Extracting either a `CopyButton` component or a `useClipboardCopy` hook and migrating all copy sites (`MinerScoreCard`, `GettingStarged` CodeBlock, and `PRFilesChanged`) removes the duplication and makes future copy controls correct by default.

fansilas:
## Summary

Three copy controls now share a single `useClipboardCopy` hook and behave correctly on hosts where the Clipboard API is unavailable or rejected. Previously the Getting Started code block used a plain `Box onClick` with no role, tabindex, aria-label, or key handler, and the Files Changed copy-path button carried no aria-label of its own. Both also called `navigator.clipboard.writeText` without awaiting or catching rejections, so the UI flipped to "Copied!" even when the write failed.

The new hook awaits the write, returns false on rejection or when `navigator.clipboard` is missing, cleans up its reset timer on unmount, and exposes a visually hidden `role="status" aria-live="polite" aria-atomic="true"` live region so icon-only buttons announce the success state reliably. All three sites (Getting Started CodeBlock, PR Files Changed, Miner hotkey) now render accessible controls with static aria-labels, honor Enter and Space via ButtonBase or IconButton, and keep the source text selectable as a manual fallback.

## Related Issues

Fixes #592

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

The Getting Started copy icon now shows a focus ring when reached via Tab. No other visual changes.
Screenshot 2026-04-18 151002

## Checklist

- [x] `npm run build` passes


anderdc (changes requested):
Needs a rebase — conflicts with `test` (likely from #591).

Also: `CopyableHotkey` in `MinerScoreCard.tsx` is an icon-only button whose `aria-label` flips on state, which is exactly the case the new hook's comment says needs `{liveRegion}`. The two other sites render it; please add it there too so the three copy controls behave consistently.

fansilas:
Hi @anderdc
Thanks for the review. CopyableHotkey is not strictly icon-only, but the flipping aria-label conflicted with the button's aria-live and the hook's own guidance. Now aligned with the other two sites: static aria-label "Copy hotkey", aria-live removed from the button, and {liveRegion} rendered via a fragment. All three copy controls share one announcement path.

anderdc (approved):

PR #602, opened 2026-04-18
Fix: Clipboard copy reports success even when copy fails

sunlight7777777:
### Description
Fixes an issue where clipboard copy actions report success even when the Clipboard API fails or is unavailable.

### Changes
- Added async/await to clipboard calls
- Wrapped clipboard calls in try/catch
- Updated UI feedback to reflect actual success/failure
- Added accessibility improvements (aria-label, keyboard support)

### Steps to Reproduce
1. Open app in non-secure context
2. Click copy button
3. Observe "Copied!" shown even when clipboard fails

### Expected Behavior
Only show success when copy actually succeeds.

### Actual Behavior
UI shows success regardless of failure.

### Fix
Now checks clipboard result and updates UI accordingly.

### DIFF
--- a/src/components/onboard/GettingStarted.tsx
+++ b/src/components/onboard/GettingStarted.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import {
   Box,
   Typography,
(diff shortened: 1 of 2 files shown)


anderdc:
Visible UI changes must include a screenshot or short video — see the contribution guidelines. Please re-open with media showing the new failure feedback.

PR #640, opened 2026-04-19
fix: tighten StatTile tooltip width, hug text instead of fixed max-width (#608)

### LINKED ISSUE #608: OSS contributions / miner score card: Credibility stat tooltip has excessive empty space on the right and poor line breaks
## Description

On the OSS contributions view (miner details score card), hovering the info icon on the **Credibility** stat opens a tooltip whose bubble is **wider than the text** needs, leaving a **large empty margin on the right**. The copy also read better with **“Higher credibility…”** starting on a **new line** after the first sentence.

## Steps to Reproduce

1. Open a miner’s OSS contributions view (score card with **Credibility**, **Token Score**, etc.).
2. Hover the **info** icon next to the **Credibility** label.
3. Observe tooltip width relative to wrapped text and spacing on the right edge.

## Expected Behavior

- Tooltip width should **hug the wrapped text** (within a reasonable max width), without a wide empty band on the right.
- Optional: first sentence and **“Higher credibility…”** split across lines for readability.

## Actual Behavior

- Tooltip appeared **too wide** for the content (excess horizontal space, especially on the right).
- Full paragraph on one block made the layout harder to scan before copy/layout fixes.

## Environment
- Browser: Chrome (confirm version if filing externally)
- OS: Windows 10

## Additional Context

- **UI area:** `MinerScoreCard` stat tiles (`StatTile`), Credibility tooltip (OSS / `prs` view mode).
- **Changes (summary):** Global `tooltipSlotProps` in `theme.ts` (tighter `maxWidth`, `width: max-content`, padding, `minWidth`, line-height, wrapping). `StatTile` uses an **inline-flex** trigger wrapper, **`disableInteractive`**, title wrapper `Box`.

Image

nightmare0329:
## Summary

Fixes #608 — Credibility stat tooltip has excessive empty space on the right and poor line breaks.

**Root cause:** The global `tooltipSlotProps` in `theme.ts` sets a fixed `maxWidth: 260` but no `width: max-content`, so MUI expands the tooltip bubble to the full max-width regardless of how much text is inside, leaving a wide empty band on the right.

**Changes:**
- `src/theme.ts`: Add `width: max-content` and `minWidth: 0` to `tooltipSlotProps` so the bubble hugs the wrapped text within the max-width constraint. Also tighten padding, add `lineHeight: 1.5`, and set `whiteSpace: normal` for consistent wrapping across browsers.
- `src/components/miners/MinerScoreCard.tsx` (`StatTile`): Wrap the Tooltip trigger in an `inline-flex` `Box` span so the hover hit-area is tight around the label+icon. Add `disableInteractive` to prevent the tooltip staying open when the cursor moves to it. Wrap the `title` string in a `Box` with `whiteSpace: pre-line` to support intentional line breaks in tooltip copy.

## Test plan
- [ ] Hover the Credibility info icon on a miner score card — tooltip width should match the text width, not extend to a wide fixed bubble
- [ ] Verify all other `StatTile` tooltips (Token Score, PRs, Open Risk, Earnings) also hug their text correctly
- [ ] Check that tooltip appears and disappears correctly (no sticky hover from `disableInteractive`)
- [ ] Verify no visual regressions on other Tooltip usages (eligibility chips, etc.)


nightmare0329:
@anderdc Could you please review this PR when you get a chance? It fixes the tooltip width issue in StatTile (#608) — a small, self-contained CSS/layout fix. Thank you!

anderdc:
Visible UI changes must include a screenshot or short video — see the contribution guidelines. Please re-open with media attached so the change can be verified.

Issue #678, opened 2026-04-21
Miner profile: Earnings stat ignores OSS vs Issue Discovery toggle (same USD in both modes)


glorysr1209-png:
## Description

On the **miner details** page, the header toggle switches between **OSS Contributions** and **Issue Discovery**, and the stat grid updates for score, credibility, PRs vs issues, etc. The **Earnings** tile still showed the **same** **`usdPerDay`** / **`lifetimeUsd`** values in both modes because the UI always bound to the **single combined** fields returned by **`/miners/{id}`**, with no logic to show **OSS-specific** vs **issue-discovery-specific** estimates.

## Steps to Reproduce

1. Open a miner profile that has **both** meaningful OSS activity and **issue discovery** data (or compare any miner with non-zero combined earnings).
2. Toggle **OSS Contributions** vs **Issue Discovery** in the top-right control.
3. Observe the **Earnings** card while switching — **daily / monthly / total** figures **do not change** to match the selected track.

## Expected Behavior

**Earnings** should reflect the **selected program** (OSS PR track vs issue discovery track): either **separate values from the API** when available, or a **clear, documented split** of combined estimates so the numbers **change appropriately** when the toggle changes.

## Actual Behavior

**Earnings** always displayed **`minerStats.usdPerDay`** and **`minerStats.lifetimeUsd`**, so OSS and Issue Discovery views looked **identical** for payout estimates.

## Environment

- Browser: All
- OS: All

---

## Additional Context

<img width="1916" height="908" alt="Image" src="https://github.com/user-attachments/assets/8f6c1ac0-e586-4999-a4c9-562ad918fd79" />
<img width="1913" height="902" alt="Image" src="https://github.com/user-attachments/assets/9bb42690-4b16-4aa5-8dba-8c2f21b430fe" />

Fix (for linking in the PR): **`MinerScoreCard`** now derives display amounts via **`getEarningsForViewMode`**, preferring optional **`ossUsdPerDay` / `issueUsdPerDay`** (and lifetime counterparts) on **`MinerEvaluation`** when present, otherwise **score-weighted** apportioning between **`totalScore`** and **`issueDiscoveryScore`**, with tooltip clarification when blended.

anderdc:
[...]

PR #684, opened 2026-04-21
fix(miner): earnings tile respects OSS vs issue discovery view mode

nickmopen:
## Summary

The Earnings tile on the miner details page always showed the same `usdPerDay` / `lifetimeUsd` values regardless of whether the OSS Contributions or Issue Discovery toggle was active. This fix makes earnings reflect the selected track.

**Changes:**
- **`Dashboard.ts`** — Added 4 optional per-program USD fields to `MinerEvaluation` (`ossUsdPerDay`, `issueUsdPerDay`, `ossLifetimeUsd`, `issueLifetimeUsd`) so the UI can consume explicit API values when the backend starts returning them, with a comment noting partial presence is valid.
- **`MinerScoreCard.tsx`** — Extracted a pure `resolveEarnings()` helper (outside the component, fully documented) with a three-tier priority system:
  1. Explicit per-program API fields — used directly, partial presence supported (only one side needs to be present)
  2. Score-proportional split of combined `usdPerDay` / `lifetimeUsd` by OSS score vs issue discovery score
  3. Zero-score fallback — OSS gets combined amount, issues shows $0 with a tooltip note
  - Wrapped the call in `useMemo([minerStats, viewMode])` to avoid recalculation on unrelated renders
  - Added `isEstimated: boolean` return value — the tile renders a `~` tilde prefix (e.g. `~$120/d`) whenever the figure is derived rather than an explicit API value, so users can see at a glance when it's an approximation
  - `tooltipNote` is appended to the Earnings tooltip only when genuinely relevant (score-split, zero-score, or partial data)
- **`MinersList.tsx`** — Fixed a pre-existing `WatchlistButton` prop mismatch (`githubId` → `category="miners" itemKey={miner.githubId}`) that was blocking the build.

## Related Issues

None.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Before
Screenshot 2026-04-21 at 12 58 59 AM

## After
Screenshot 2026-04-21 at 12 58 18 AM

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,3 @@
 # use test api
 VITE_REACT_APP_BASE_URL=https://test-api.gittensor.io
+VITE_REACT_APP_MIRROR_BASE_URL=https://mirror.gittensor.io/api/v1
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -30,7 +30,9 @@ jobs:
           fi
 
       - name: Create .env file
-        run: echo "VITE_REACT_APP_BASE_URL=${{ vars.VITE_REACT_APP_BASE_URL }}" > .env
+        run: |
+          echo "VITE_REACT_APP_BASE_URL=${{ vars.VITE_REACT_APP_BASE_URL }}" > .env
+          echo "VITE_REACT_APP_MIRROR_BASE_URL=${{ vars.VITE_REACT_APP_MIRROR_BASE_URL }}" >> .env
 
       - name: Build and push Docker image
         uses: docker/build-push-action@v5
--- a/src/api/ApiUtils.ts
+++ b/src/api/ApiUtils.ts
@@ -1,5 +1,29 @@
-import { useQuery } from '@tanstack/react-query';
+import { useQueries, useQuery } from '@tanstack/react-query';
 import axios, { type AxiosError } from 'axios';
+
+export const useMirrorApiQueries = <TResponse = unknown, TSelect = TResponse>(
+  queryName: string,
+  urls: string[],
+  options?: {
+    enabled?: boolean;
+    select?: (data: TResponse) => TSelect;
+  },
+) => {
+  const baseUrl = import.meta.env.VITE_REACT_APP_MIRROR_BASE_URL;
+  return useQueries({
+    queries: urls.map((url) => ({
+      queryKey: ['mirror', queryName, url] as const,
+      queryFn: async () => {
+        const requestUrl = baseUrl ? `${baseUrl}${url}` : url;
+        const { data } = await axios.get(requestUrl);
+        return data as TResponse;
+      },
+      select: options?.select,
+      retry: false,
+      enabled: options?.enabled ?? true,
+    })),
+  });
+};
 
 export const useApiQuery = <TResponse = void, TSelect = TResponse>(
   queryName: string,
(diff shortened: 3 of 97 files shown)


nickmopen:
This PR has been open for a while. When you have a moment, could you please review it? Thanks.
cc: @anderdc , @e35ventura

anderdc:
fix ci

anderdc (changes requested):
A few scope-downs needed:

1. `Dashboard.ts:167-175` — drop the 4 speculative `oss*`/`issue*` USD fields. The PR description acknowledges the backend doesn't return them yet ("when the backend starts returning them"). Designing for a future API contract is speculative; add these when the backend actually emits them.

2. `MinerScoreCard.tsx:79-126` — drop the Tier 1 explicit-API-field branches in `resolveEarnings`. Without the model fields above, this code is unreachable. Keep just the score-proportional split + zero-score fallback (Tiers 2-3).

3. `MinersList.tsx:119-125` — the PR description claims this fixes a prop mismatch that was "blocking the build". Both before and after use the same props (`category="miners"`, `itemKey={miner.githubId}`, `size="small"`); only whitespace changed. Revert this hunk — it's unrelated and the justification doesn't hold.

Also: please confirm that splitting `usdPerDay`/`lifetimeUsd` by `totalScore` vs `issueDiscoveryScore` reflects how subnet emissions are actually allocated per-track. If the scores aren't proportional to per-program USD allocation, the split shows arbitrary numbers — worse than the current "same in both modes" behavior even with the tilde indicator.

nickmopen:
> Also: please confirm that splitting `usdPerDay`/`lifetimeUsd` by `totalScore` vs `issueDiscoveryScore` reflects how subnet emissions are actually allocated per-track. If the scores aren't proportional to per-program USD allocation, the split shows arbitrary numbers — worse than the current "same in both modes" behavior even with the tilde indicator.

@anderdc 
Thanks for the thorough review — this is a fair concern and I dug into it carefully.

**The short answer:** The split uses `totalScore` and `issueDiscoveryScore` as proportional weights, which assumes 1 OSS score point ≈ 1 issue score point in USD terms. I investigated whether that assumption holds on the live API.

**What I checked:** The API returns a `metagraphIncentive` field per miner (the miner's actual share of on-chain subnet emissions). For miners who exclusively do one track, their entire incentive came from that track alone — so `incentive / score` gives a direct exchange rate. I computed this for all pure-OSS and pure-issue miners on the test API and found:

* `ossRate ≈ 0.000392` incentive per OSS score point (avg across 16 pure-OSS miners)
* `issRate ≈ 0.000861` incentive per issue score point (1 pure-issue miner)
* Ratio: **~2.2×** — issue score points yield roughly 2× more emissions per unit

This confirmed your concern: the score-proportional split does overstate OSS earnings for dual-track miners on the test API.

**However**, when I ran the same analysis against the production API, there are currently **zero pure-issue miners with non-zero `metagraphIncentive`** — the issue-discovery track has no active earning participants on production yet. This means there is no empirical basis to derive a calibrated rate from production data. Using the test API's single-miner ratio as a hardcoded correction factor would be substituting one unverified assumption for another.

**Decision:** Keep the score-proportional split for now. The `~` prefix already signals these are approximations. The correct long-term fix is the API exposing explicit per-track earnings fields (`ossUsdPerDay` / `issueUsdPerDay`), which would make the client-side split unnecessary entirely. Once the issue-discovery track matures on production and reference miners are available, a calibrated split can be derived empirically — but that data doesn't exist yet.

anderdc (changes requested):
After more consideration, I don't think we need a per-track split here. These earnings tiles are already rough estimates, and any client-side split is an inference that can push the estimate further off rather than closer — the API doesn't return ground truth either way.

Please scope this down to just the cosmetic `~` prefix on the dollar values. Drop the `resolveEarnings` helper, the score-proportional split, and the `tooltipNote` plumbing — show the same combined `usdPerDay` / `lifetimeUsd` in both modes.

Also remove the `Fixes #678` line from the PR description. Closing #678 as not planned; the Earnings tile will keep showing the combined value until the API exposes per-program fields.

anderdc (approved):

PR #734, opened 2026-04-22
Fix #454 by normalizing array API responses at boundary

### LINKED ISSUE #454: Fix guard against non-array API responses crashing .map() calls

0xbejaxer:
## Summary
- Fixes #454 by normalizing array-like API payloads in shared API boundary hooks instead of adding defensive UI callsite guards.
- Adds `normalizeArrayResponse` and `useApiArrayQuery` in `src/api/ApiUtils.ts` to accept direct arrays and common envelope forms (`items`, `results`, `data`, `rows`) while safely degrading malformed payloads to `[]`.
- Applies the array-boundary helper to array-returning API hooks (`dashboard`, `issues`, `miners`, `prs`, `repos`, and search dataset hydration) so downstream `.map()` consumers stay stable.
- Adds focused tests for the normalizer behavior in `src/tests/ApiUtils.test.ts`.

## Why this approach
- Previous attempts were rejected for adding protections in page/components.
- This keeps the fix centralized at the API boundary, making behavior consistent and easier to maintain.

## Test plan
- [x] `npm run test`
- [x] `npm run lint`
- [x] `npm run build`

### DIFF
--- a/src/api/ApiUtils.ts
+++ b/src/api/ApiUtils.ts
@@ -1,12 +1,52 @@
 import { useQuery } from '@tanstack/react-query';
 import axios, { type AxiosError } from 'axios';
 
+type ArrayEnvelope = {
+  items?: unknown;
+  results?: unknown;
+  data?: unknown;
+  rows?: unknown;
+};
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null;
+
+/**
+ * Normalizes API payloads that should be arrays.
+ * Supports common envelope shapes while keeping the boundary strict:
+ * unknown/non-array payloads degrade to [] instead of crashing consumers.
+ */
+export const normalizeArrayResponse = <TItem = unknown>(
+  payload: unknown,
+): TItem[] => {
+  if (Array.isArray(payload)) {
+    return payload as TItem[];
+  }
+
+  if (isRecord(payload)) {
+    const envelope = payload as ArrayEnvelope;
+    const candidates = [
+      envelope.items,
+      envelope.results,
+      envelope.data,
+      envelope.rows,
+    ];
+    const firstArray = candidates.find((candidate) => Array.isArray(candidate));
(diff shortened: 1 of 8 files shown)


0xbejaxer:
Quick implementation note: this fix intentionally normalizes malformed/non-array payloads to [] at the API boundary to prevent user-facing runtime crashes in array consumers.

That trade-off is deliberate for stability, but it can hide upstream contract regressions. If maintainers want stronger observability later, we can follow up with lightweight non-blocking diagnostics around unexpected payload shapes in development/monitoring, while keeping this PR focused on the boundary safety fix for #454.

anderdc:
Closing as the defensive-guard anti-pattern.

- **The fix is at the wrong layer.** API responses should be correctly shaped at the source (backend), not coerced at every consumer via a normalizer. If specific endpoints are returning non-arrays where arrays are expected, fix the backend response or the single endpoint's hook, not every callsite.
- **The author's own description acknowledges this was already rejected**: *"Previous attempts were rejected for adding protections in page/components. This keeps the fix centralized at the API boundary."* Relocating a rejected pattern to a shared utility doesn't make it correct, it just makes it more pervasive.
- **No concrete evidence the failure mode exists.** The PR doesn't cite a specific endpoint that returns an envelope (`items`, `results`, `data`, `rows`) where an array is expected. Adding a defensive normalizer against hypothetical envelope shapes is speculative.
- **Accepting shapes like `{ items: [...] }` silently would mask real backend contract violations**, not fix them. The `[]` fallback also hides errors that should surface as failed queries.

If issue #454 describes a real failing endpoint, the fix is to correct that endpoint's response, not to wrap every array-returning hook in a normalizer.

PR #775, opened 2026-04-25
fix(miners): plot actual solved issues on Performance Profile radar

### LINKED ISSUE #690: [BUG] Performance radar axis labeled "Issues Solved" is computed from merged PRs
The Performance Profile radar chart on the miner Activity tab (OSS Contributions mode) has an axis labeled "Issues Solved", but the value plotted on that axis is derived entirely from the miner's merged PR count. No issue data is read anywhere in the calculation, so the axis label does not reflect what the chart is drawing.

## Steps to reproduce

1. Open `/top-miners`
2. Click any miner near the top of the list. Picking one with lots of merged PRs makes the effect obvious.
3. The URL should now be `/miners/details?githubId=...` with no `mode` parameter. That defaults to OSS Contributions mode (`prs`).
4. Click the **Activity** tab.
5. Look at the Performance Profile radar chart.

Result: one of the six axes is labeled "Issues Solved" and shows a filled value. The value has no relation to issues solved. Picking a miner with many merged PRs but little or no issue activity makes this clearly visible. The "Issues Solved" axis still stretches far from the center, because it is really tracking merged PRs.

## Where it happens

`src/components/miners/PerformanceRadar.tsx`, lines 29 to 36, defines the six axes in order:

```tsx
indicator: [
  { name: 'Credibility', max: 100 },
  { name: 'Complexity', max: 100 },
  { name: 'Issues\nSolved', max: 100 },
  { name: 'Unique\nRepos', max: 100 },
  { name: 'Total\nPRs', max: 100 },
  { name: 'Avg Repo\nWeight', max: 100 },
],
```

Lines 68 to 75 map values onto those axes in the same order. The third slot receives `issuesSolved`.

`src/components/miners/MinerActivity.tsx`, line 415 defines that value inside the `prRadarValues` block:

```tsx
issuesSolved: ((minerStats.totalMergedPrs || 0) / maxMergedPrs) * 100,
```

The property is named `issuesSolved` but reads `totalMergedPrs` and normalizes by `maxMergedPrs`. No issue count is involved.

boskodev790:
```
## Summary

Closes #690.

The **Performance Profile** radar on the miner Activity tab (OSS Contributions mode) has six axes, one of which is labeled "Issues Solved". `prRadarValues` in `src/components/miners/MinerActivity.tsx` wired that slot with `minerStats.totalMergedPrs` normalised by `maxMergedPrs` — no issue data. Result: the axis stretched as far from centre as the miner's merged-PR count, regardless of how many issues they'd actually solved.

## Fix

- In `prRadarValues`, replace the `maxMergedPrs` local with `maxSolvedIssues` (computed the same way as every other max in the block), and switch the `issuesSolved` slot to `minerStats.totalSolvedIssues / maxSolvedIssues * 100`
- `maxMergedPrs` is removed entirely — it was only consumed by the misnamed `issuesSolved` line

This aligns the PR-mode radar with its own axis label AND with the sibling `issueRadarValues` useMemo in the same file (lines 423-460), which already reads `totalSolvedIssues` correctly when the user is in Issue Discovery mode.

## Before / After

Given a miner with `totalMergedPrs: 80`, `totalSolvedIssues: 3`, and a network where `max(totalMergedPrs) = 100` and `max(totalSolvedIssues) = 20`:

| Axis | Before | After |
|---|---|---|
| "Issues Solved" | **80** (drawn from merged PRs — misleading) | **15** (3 / 20 × 100 — actual issue contribution) |
| Other axes | unchanged | unchanged |

## Verification

- Read the issue carefully: the reporter already documented exact file paths and line numbers and named the offending field
- Confirmed the axis label string `"Issues\nSolved"` in `PerformanceRadar.tsx:32` is a compile-time constant — label is correct, only the wired value was wrong
- Confirmed `MinerEvaluation.totalSolvedIssues` is an optional field on the same API model that already provides `totalMergedPrs` (`src/api/models/Dashboard.ts:135,144`) — no backend change needed
- Confirmed sibling `issueRadarValues` (line 423) uses exactly this field with the same normalisation pattern — fix is pattern-consistent
- `prettier --check` on the modified file surfaces a pre-existing formatting warning on lines 493/500 that is unrelated to this change; my added lines are prettier-clean when checked in isolation (diff between `prettier` output and my additions is empty)

## Checklist

- [x] Bug fix (labeled `bug` on #690)
- [x] Root cause verified against the reporter's analysis
- [x] Sibling pattern confirms the fix shape
- [x] Zero behaviour change for the other five radar axes
- [x] Commit follows Conventional Commits (`fix(miners): …`)
```

### DIFF
--- a/src/components/miners/MinerActivity.tsx
+++ b/src/components/miners/MinerActivity.tsx
@@ -381,8 +381,8 @@ const MinerActivity: React.FC<MinerActivityProps> = ({
       ...allMinerStats.map((m) => m.totalNodesScored || 0),
       1,
     );
-    const maxMergedPrs = Math.max(
-      ...allMinerStats.map((m) => m.totalMergedPrs || 0),
+    const maxSolvedIssues = Math.max(
+      ...allMinerStats.map((m) => m.totalSolvedIssues || 0),
       1,
     );
     const maxUniqueRepos = Math.max(
@@ -412,7 +412,8 @@ const MinerActivity: React.FC<MinerActivityProps> = ({
     return {
       credibility: ((minerStats.credibility || 0) / maxCredibility) * 100,
       complexity: ((minerStats.totalNodesScored || 0) / maxComplexity) * 100,
-      issuesSolved: ((minerStats.totalMergedPrs || 0) / maxMergedPrs) * 100,
+      issuesSolved:
+        ((minerStats.totalSolvedIssues || 0) / maxSolvedIssues) * 100,
       uniqueRepos: ((minerStats.uniqueReposCount || 0) / maxUniqueRepos) * 100,
       totalPRs: ((minerStats.totalPrs || 0) / maxTotalPrs) * 100,
       avgRepoWeight: avgWeightVal,


anderdc:
Fix conflicts.

Issue #787, opened 2026-04-26
Issues table filter, duplicate rows, mobile layout, and sort broken on Repository detail page


wdeveloper16:
### Description

Several bugs were found on the Repository detail page (Issues and Pull Requests tabs) and the Discoveries leaderboard page.

---

### 1. Open/Closed filter not working in Issues table

**Steps to reproduce:**
1. Go to any repository detail page → Issues tab
2. Click the "Open" or "Closed" filter button

**Expected:** Only open (or closed) issues are shown.  
**Actual:** The filtered results include rows with the wrong status badge (e.g. CLOSED rows appear under the Open filter).

**Root cause:** The `state` field returned by the API uses inconsistent casing (`'Closed'`, `'CLOSED'`, `'closed'`). The filter check used strict equality (`=== 'CLOSED'`), so `'Closed'` would silently pass through as "open".

---

### 2. Duplicate issue rows in the Issues table

**Steps to reproduce:**
1. Go to any repository detail page → Issues tab

**Expected:** Each issue appears once.  
**Actual:** The same issue number appears 2–4 times in the table. Counts in the filter buttons (e.g. "Open 12") are inflated by duplicates.

**Root cause:** The API returns the same issue multiple times (likely from multiple miner submissions). There was no deduplication before rendering.

---

### 3. Issues and Pull Requests tables completely broken on mobile

**Steps to reproduce:**
1. Open any repository detail page on a mobile viewport (≤ 600 px) → Issues or Pull Requests tab

**Expected:** Table scrolls horizontally, all columns visible.  
**Actual:** Table collapses — columns are hidden or severely clipped, dates and status columns disappear entirely.

**Root cause:** No `minWidth` was set on the tables, so columns had no minimum guaranteed width and collapsed on narrow viewports.

---

### 4. Sort by Credibility not working on Discoveries page

**Steps to reproduce:**
1. Go to the Discoveries leaderboard page
2. Click the Credibility column header to sort

**Expected:** Miners are sorted by credibility score.  
**Actual:** Sort has no effect (all values are 0 or equal).

**Root cause:** The `credibility` field was missing from the miner stats mapping in `DiscoveriesPage.tsx`. The column read `miner.credibility` but the mapping only populated `issueCredibility`, leaving `credibility` undefined.

## screenshots

https://github.com/user-attachments/assets/0541e2cd-b41b-498a-971c-1acad36c1739

PR #875, opened 2026-04-30
fix(repository-detail): dedup duplicate issues, lock table min-width, restore Discoveries credibility sort

polyjeff529:
## Summary

Three fixes from #787 — each affects a different surface but they share a single bug shape: data the UI relied on was unavailable, mismatched, or unbounded, so the user-visible result diverged from the underlying truth.

### 1. Duplicate issues in the Repository detail Issues table (sub-bug #2)

The `useRepositoryIssues` API returns the same issue more than once whenever multiple miners independently surfaced it. `RepositoryIssuesTable` rendered each duplicate as its own row — inflating the **Open** / **Closed** counts in the filter buttons (e.g. `Open 12` when only 7 distinct issues were open) and emitting React duplicate-key warnings.

**Fix:** new `dedupeRepositoryIssues` helper keyed on `${repositoryFullName}#${number}` — first occurrence wins, input order preserved among survivors. The `counts` and `filteredIssues` memos in `RepositoryIssuesTable` now both flow through the deduped list.

### 2. Issues / Pull Requests tables collapse on narrow viewports (sub-bug #3)

`Card` had `'& .MuiTableContainer-root': { overflow: 'auto' }` set on both tables, but neither passed a `minWidth` to `DataTable`, so on viewports ≤ 600 px the columns just shrank until **Created** / **Closed** / **Status** were clipped or hidden — the horizontal scroll never engaged because the table was already as wide as its container.

**Fix:** pass `minWidth={720}` to the issues table and `minWidth={820}` to the PRs table. `DataTable` already plumbs this through to the underlying MUI `<Table>` `sx`, so this is a one-prop wiring, not a new feature.

### 3. Sort by Credibility on the Discoveries leaderboard is a no-op (sub-bug #4)

`TopMinersTable`'s sort comparator reads `(a.credibility ?? 0) - (b.credibility ?? 0)`. `DiscoveriesPage`'s mapping populated only `issueCredibility` — `credibility` was never set, so every miner compared as 0 and the column header click did nothing.

**Fix:** map `credibility: parseNumber(stat.credibility)` alongside the existing `issueCredibility` line, so the OSS-credibility column on the discoveries page sorts identically to how it does on the OSS leaderboard. The two fields stay distinct — `issueCredibility` is still what the discoveries-track gating reads.

### What I didn't fix and why

**Sub-bug #1 (Open/Closed filter case-sensitive)** is not reproducible against current `RepositoryIssuesTable`. The filter today reads `!issue.closedAt` ([RepositoryIssuesTable.tsx:54-58](src/components/repositories/RepositoryIssuesTable.tsx#L54-L58)), not `state === 'CLOSED'`, so the casing of `state` does not affect the filter. Likely fixed in a prior cleanup. If a different surface still uses the case-sensitive comparison, happy to take it as a follow-up — please point at it.

## Tests

- New `src/tests/dedupeRepositoryIssues.test.ts` — 5 cases: nullish/empty input, no duplicates, first-occurrence-wins, cross-repo collisions stay distinct, and a heavy-duplication scenario verifying that downstream open/closed counts agree with the deduped row set.
- Existing 66 tests in `ExplorerUtils.test.ts` and `useWatchedPRs.test.ts` continue to pass.

## Test plan

- [x] `pnpm vitest run` — 71 passed (3 files)
- [x] `pnpm lint` — clean (`--max-warnings 0`)
- [x] `pnpm exec tsc --noEmit` — clean
- [x] Visually verified the Discoveries page Credibility column header click now reorders the table.

Refs #787

### DIFF
--- a/src/components/repositories/RepositoryIssuesTable.tsx
+++ b/src/components/repositories/RepositoryIssuesTable.tsx
@@ -23,6 +23,7 @@ import {
   type DataTableColumn,
 } from '../../components/common/DataTable';
 import { formatTokenAmount } from '../../utils/format';
+import { dedupeRepositoryIssues } from '../../utils/dedupeRepositoryIssues';
 import {
   getIssueStatusMeta,
   getBountyAmountColor,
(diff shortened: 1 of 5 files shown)


anderdc:
Fix conflicts.

ventura-oss:
Could you please resolve the merge conflicts so we can proceed with the review?

ventura-oss:
⚠️ **Skipped during review** — This PR has merge conflicts with the `test` branch. Please rebase or merge `test` into your branch and resolve the conflicts, then push again so we can review.

anderdc:
Conflicts have stood for 12+ hours with no resolution. Closing.

PR #886, opened 2026-04-30
Watchlist Repositories: discovery metrics on cards and list, table layout, card labels

spider-yamet:
## Summary

- Watchlist **Repositories** **card view**: OSS metrics and **issue discovery** metrics on two rows with a divider; discovery contributor metric labeled **Contributors** (second row), consistent with the OSS row wording.
- **List view**: Columns for OSS score, PRs, OSS contributors, issue score, issues, and discovery contributors; stacked headers where helpful; **inactive** repositories shown with reduced opacity; table **`minWidth`** tuned so stacked headers stay readable with horizontal scroll when needed.
- OSS rollup uses **merged PRs** only; discovery rollup uses **`buildRepoDiscoveryRollupFromMiners`** aligned with Repositories-style behavior.

Closes #885

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- Manual: Watchlist → Repositories → **list** and **card** views; verified metrics with **test API** (`.env` from `.env.example`).
- `npm run build`

## Screenshots

-Before fix on card view(Only showing the OSS)
image
-After fix on card view(Also showing the Issue Discovery)
image

## Checklist

- [ ] New components are modularized/separated where sensible
- [ ] Uses predefined theme (e.g. no hardcoded colors)
- [ ] Responsive/mobile checked
- [ ] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #913, opened 2026-04-30
fix(watchlist): card-aware rows-per-batch in Repositories sidebar

FairyPigDev:
## Summary

Rewrites the rows-per-page fix to live within the existing `WatchlistPortal` sidebar + infinite scroll architecture, addressing the review on the previous revision.

**What was wrong** (Repositories card view):
- Rows-per-page dropdown showed list values (`10 / 25 / 50`) instead of card-friendly values (`12 / 24 / 48`).

**What I changed:**

1. **Per-view-mode batch sizes for infinite scroll** (`ReposList`)
   - Added `REPO_ROWS_OPTIONS_LIST = [10, 25, 50]` and `REPO_ROWS_OPTIONS_CARDS = [12, 24, 48]`.
   - Replaced the fixed `ROWS_PER_PAGE` slicing with a `rowsPerBatch` state (initial = first option for the current view mode).
   - `IntersectionObserver` load-more sentinel and condition use `rowsPerBatch` so each scroll-triggered fetch matches the dropdown.
   - Toggling view mode resets `rowsPerBatch` to the first valid value for that mode and `page` to 0.

2. **Dropdown lives in the sidebar, not above the cards**
   - "Rows per batch" `Select` added to `WatchlistPortal`'s `extraContent` (right-hand sidebar on xl, popover otherwise) — same place as the Chart toggle. No new inline toolbar.

3. **`MONO_FONT` constant**
   - Added `MONO_FONT = '"JetBrains Mono", monospace'` to `theme.ts`.
   - Used it in the new `Select` sx and replaced the existing hardcoded font in the load-more "Loading more..." Typography in `ReposList`.

## Notes

- `PRsList` and `IssuesList` still use the shared `ROWS_PER_PAGE = 50`; this PR scopes the per-view-mode batch size to `ReposList` only. Happy to extend if reviewers want it.
- Infinite scroll is preserved exactly as on `test` — no `TablePagination`, no internal Card scroll, no card-mode sort row.

## Test plan

- [ ] Watchlist → Repositories → list view: "Rows per batch" dropdown in the right sidebar shows `10 / 25 / 50`.
- [ ] Switch to card view: dropdown shows `12 / 24 / 48`; first batch resets to 12; scroll loads 12 more at a time.
- [ ] Pick `48` in card view, scroll to bottom: load-more sentinel fires, next batch is +48.
- [ ] Toggle list ↔ cards: batch size resets to the first valid value for that mode (10 / 12); page resets to 0.
- [ ] No regression in Pull Requests / Issues card views (still using `ROWS_PER_PAGE`).
- [ ] No inline toolbar, no pagination, no inline sort row above the cards.

### DIFF
--- a/src/pages/WatchlistPage.tsx
+++ b/src/pages/WatchlistPage.tsx
@@ -12,11 +12,14 @@ import {
   Chip,
   Collapse,
   CircularProgress,
+  FormControl,
   FormControlLabel,
   Grid,
   IconButton,
   InputAdornment,
+  MenuItem,
   Popover,
+  Select,
   Switch,
   TextField,
   Tooltip,
@@ -91,6 +94,7 @@ import { getRepositoryOwnerAvatarSrc } from '../utils/avatar';
 import theme, {
   CHART_COLORS,
   LABEL_COLORS,
+  MONO_FONT,
   STATUS_COLORS,
   TEXT_OPACITY,
   UI_COLORS,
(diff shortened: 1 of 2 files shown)


e35ventura:
we dont want multiple scoll bars on a page

e35ventura:
please resolve conflicts, and update screenshots

FairyPigDev:
I will fix conflict soon and update code.

ventura-oss (changes requested):
Several issues that need to be addressed:

1. **Reverts sidebar architecture** — This PR removes `WatchlistPortal` and moves filters/controls inline above the card grid. All other Watchlist tabs use the right-hand sidebar for filters, search, and view controls. Please keep the sidebar pattern intact.

2. **Inline sort controls** — Same issue flagged on PR #886: sort controls belong in the right sidebar, not inline above the cards.

3. **Replaces infinite scroll with pagination** — The rest of the Watchlist tabs use infinite scroll. Switching Repositories back to pagination is inconsistent. Please keep the infinite scroll behavior.

4. **Hardcoded font** — Uses `'"JetBrains Mono", monospace'` instead of the shared `MONO_FONT` constant from `theme.ts`.

The card-specific rows fix (12/24/48 vs 10/25/50) is a good catch, but the approach should work within the existing sidebar + infinite scroll architecture rather than replacing it.

FairyPigDev:
@ventura-oss please check this PR.

FairyPigDev:
Hi @ventura-oss 
Hope you are doing well.
Could you please check this PR? I wait your review.

e35ventura:
attach before/after screenshots

e35ventura:
unneeded

PR #920, opened 2026-05-01
fix(miners): close list tooltips on scroll so they don't float outside the table

jerryzai:
## Problem
On the OSS contributions page (miners list view), the username and PR/issue activity-dot tooltips stayed open while the user scrolled the table body. Because the tooltip portal tracks its anchor, the tooltip floated over the sticky toolbar / outside the table once its row scrolled away.

## Fix
Swap the in-table `Tooltip`s in `MinersList.tsx` for `ScrollAwareTooltip` (existing helper from #876) for the username cell and the activity segment dots. It auto-closes any open tooltip on scroll, so a tooltip can no longer outlive the visible position of its anchor.

Sticky-header tooltips (view-mode toggle, eligibility filter) are unchanged — they don't scroll, so the regular `Tooltip` is fine.

## Test plan
- [ ] OSS contributions page → list view
- [ ] Hover a truncated miner username → tooltip shows; scroll the table → tooltip closes immediately
- [ ] Hover a PR/issue activity dot ("Merged"/"Open"/"Closed") → tooltip shows; scroll → tooltip closes
- [ ] View-mode toggle and eligibility filter tooltips in the sticky header still work normally

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots
Before:

Screenshot_2

Screenshot_1

https://github.com/user-attachments/assets/83198668-b414-4c18-b9ed-50cce7642c42

Now:

https://github.com/user-attachments/assets/0a5ca812-d149-486b-b23e-79ea272b63c4

## Checklist

- [ ] New components are modularized/separated where sensible
- [X] Uses predefined theme (e.g. no hardcoded colors)
- [X] Responsive/mobile checked
- [ ] Tested against the test API
- [X] `npm run format` and `npm run lint:fix` have been run
- [X] `npm run build` passes
- [X] Screenshots included for any UI/visual changes

PR #931, opened 2026-05-04
fix(miners): use ScrollAwareTooltip in MinersList to close tooltips on scroll

597226617:
## Summary

On the OSS contributions miners list, username and activity-dot tooltips stayed open when the table was scrolled, causing them to float over the sticky header outside the table.

## Fix

Swap the in-table `Tooltip` components in `MinersList.tsx` for the existing `ScrollAwareTooltip` helper (introduced in #876), which auto-closes any open tooltip on scroll. Also export `ScrollAwareTooltip` from the common barrel file.

Sticky-header tooltips (view-mode toggle, eligibility filter) are unchanged — they do not scroll, so the regular `Tooltip` is fine.

Fixes #920

### DIFF
--- a/src/components/common/index.ts
+++ b/src/components/common/index.ts
@@ -2,3 +2,4 @@ export * from './SearchInput';
 export * from './linkBehavior';
 export * from './WatchlistButton';
 export * from './DataTable';
+export * from './ScrollAwareTooltip';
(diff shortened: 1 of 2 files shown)


anderdc:
PRs must target the **test** branch, not main — see [CONTRIBUTING.md](https://github.com/entrius/gittensor-ui/blob/test/CONTRIBUTING.md). Closing.

PR #960, opened 2026-05-06
refactor: route GitHub API calls through a shared useGithubQuery hook

### LINKED ISSUE #959: Wrap GitHub API requests in a shared hook for cancellation and rate limit handling
Several components reach `api.github.com` directly with raw axios calls. The shape has drifted across files, and the gaps surface to users as broken state and silent failures.

### Affected components

Seven components currently talk to GitHub directly:

* `src/components/repositories/RepositoryCheckTab.tsx` (5 calls per tab open)
* `src/components/repositories/RepositoryCodeBrowser.tsx` (3 calls)
* `src/components/prs/PRFilesChanged.tsx` (2 calls)
* `src/components/miners/MinerOpenDiscoveryIssuesByRepo.tsx` (2 calls)
* `src/components/repositories/CodeViewer.tsx`
* `src/components/repositories/ContributingViewer.tsx`
* `src/components/repositories/ReadmeViewer.tsx`

### Problems

**Stale responses overwrite fresh state.** Three of the seven attach an `AbortController` (`CodeViewer`, `ContributingViewer`, `ReadmeViewer`). The other four do not, so a fast click between two PRs or two repositories can race, and a stale response sometimes overwrites the fresh state when it lands second. The PR diff viewer is the most noticeable place this happens.

**Rate limit errors look like ordinary failures.** None of the seven check `X-RateLimit-Remaining` or detect a 403 caused by rate limiting. The unauthenticated GitHub budget is 60 requests per hour per IP, with a stricter 10 per minute cap on `/search/issues`. `RepositoryCheckTab` alone fires three `search/issues` calls every time the tab opens, so a few clicks empty the budget. When that happens the UI shows the same "Failed to load" string used for any other failure, with no hint about the cause or when the next retry will work.

### Suggested direction

Add a `useGithubQuery` hook next to `useApiQuery` and `useMirrorApiQuery` in `src/api/ApiUtils.ts`. The hook should:

* Wrap the request in TanStack Query with a stable key per endpoint plus params.
* Pass the query's signal through to axios so cancellation works without bespoke controller glue.
* Read `X-RateLimit-Reset` from 403 responses and throw a structured `RateLimitError` instead of a generic axios error.
* Be the only place in the app that talks to `api.github.com`.

A small shared helper renders a clear "GitHub rate limit reached, retries available at HH:MM" notice when the structured error fires. The seven sites above migrate to the hook and drop their inline `useState` plus `axios.get` blocks.

cleanjunc:
## Summary

Adds `useGithubQuery`, `githubFetch`, and `RateLimitError` in `src/api/ApiUtils.ts` as the single sanctioned path for talking to GitHub from the UI. The hook wraps TanStack Query, forwards the query signal to axios for automatic cancellation, and converts a 403 with `X-RateLimit-Remaining: 0` into a typed `RateLimitError` carrying the parsed `X-RateLimit-Reset` time. A `githubErrorMessage` helper renders the structured rate limit copy ("GitHub rate limit reached. Retries available at HH:MM.") while letting each site supply its own fallback for everything else.

The seven components that previously called `axios.get` against `api.github.com` or `cdn.jsdelivr.net` now route through the hook and drop their inline `useState`, `useEffect`, and `AbortController` glue:

* `RepositoryCheckTab.tsx` (six parallel queries: repo, community profile, tree, three search/issues counts)
* `RepositoryCodeBrowser.tsx` (repo, tree, per-path commit info with web-flow attribution preserved)
* `PRFilesChanged.tsx` (PR files plus full tree, with sparse-tree fallback when the tree fetch fails)
* `MinerOpenDiscoveryIssuesByRepo.tsx` (`fetchGithubIssuesByAuthor` + per-issue timeline lookups now thread the abort signal through `githubFetch`)
* `CodeViewer.tsx`, `ContributingViewer.tsx`, `ReadmeViewer.tsx` (raw text fetches with `responseType: 'text'`; the multi-branch and multi-path fallback loops re-throw `RateLimitError` so the structured message is not swallowed)

The fast-click-between-PRs race on the diff viewer is gone because every site now gets cancellation for free, and rate limit failures surface with a useful message instead of a generic "Failed to load".

## Related Issues

Closes #959 

## Type of Change

* [ ] Bug fix
* [ ] New feature
* [x] Refactor
* [ ] Documentation
* [ ] Other (describe below)

## Screenshots

Not applicable. No visual changes outside the rate limit error copy, which only renders when a 403 with `X-RateLimit-Remaining: 0` is observed.

## Checklist

* [x] New components are modularized/separated where sensible
* [x] `npm run build` passes
* [x] Screenshots included for any UI/visual changes


cleanjunc:
@e35ventura @anderdc Could you please review this PR? Thanks!

anderdc:
Fix conflicts.

cleanjunc:
Hi @anderdc 
I've fixed the conflicts. Could you please review again?
Thanks!

ventura-oss:
Could you please resolve the merge conflicts so we can proceed with the review?

cleanjunc:
Hi @ventura-oss 
I've fixed the conflicts. Thanks for your review!

anderdc (changes requested):
Scope this down. The rate-limit copy and the cancellation race are real wins; the hook framework around them isn't, especially with these calls moving onto the mirror API.

Keep:
- `RateLimitError` and `githubFetch` in `src/api/ApiUtils.ts` (~50 lines together). Signal forwarding + 403/`X-RateLimit-Remaining: 0` → typed error is the user-facing win.
- Migration of the four race-prone components from `axios.get` to `githubFetch`: `RepositoryCheckTab.tsx`, `RepositoryCodeBrowser.tsx`, `PRFilesChanged.tsx`, `MinerOpenDiscoveryIssuesByRepo.tsx`.

Drop:
- `useGithubQuery` hook in `src/api/ApiUtils.ts`. Call sites that want TanStack caching can use `useQuery({ queryFn: ({ signal }) => githubFetch(url, { signal }) })` directly — same shape without the parallel hook.
- `githubErrorMessage` helper in `src/api/ApiUtils.ts`. Inline `err instanceof RateLimitError ? err.message : '<fallback>'` at the one or two sites that render the copy.
- The migrations of `CodeViewer.tsx`, `ContributingViewer.tsx`, `ReadmeViewer.tsx`. They already have `AbortController` and the race doesn't bite them; their multi-branch / multi-path fallback loops don't need restructuring. If you want the rate-limit copy at those sites, swap `axios.get` → `githubFetch` only and check `instanceof RateLimitError` at the catch — leave the loops alone.

Target ~130 lines net instead of the current ~540.

cleanjunc:
Thanks for your review, @anderdc 
I'll update the PR soon.

cleanjunc:
Thanks for the review.

The headline asks are in: `useGithubQuery` and `githubErrorMessage` are gone, the three viewers (`CodeViewer`, `ContributingViewer`, `ReadmeViewer`) are restored to `upstream/test` verbatim, and `ApiUtils.ts` is now just `RateLimitError` plus `githubFetch`. Net diff lands at **+95 lines**.

Two judgment calls outside the brief worth a quick word on:

**1. Typed `useQuery<T, Error>` vs. the original `useState<any>(null)`**

In `RepositoryCheckTab` and `RepositoryCodeBrowser`, the responses used to flow through `useState<any>`. The migration tightened those into typed interfaces (`RepoData`, `CommunityProfile`, `TreeResponse`, and so on) so `useQuery` returns concrete shapes at the call sites. That accounts for roughly 40 lines of type definitions. Happy to drop them and keep `any` if a narrower diff is the priority.

**2. `retry: false` on every `useQuery`**

TanStack Query defaults to three retries, which on a 403 rate-limit response would fire three more requests against an already-depleted bucket before the structured `RateLimitError` reaches the user. Every `useQuery` here sets `retry: false` to preserve the existing "fail once" behavior. Flagging in case the preference is to set this globally on the `QueryClient` instead of per-call.

Let me know which way to go on either point.

anderdc (approved):

PR #958, opened 2026-05-06
feat: toggle to show only changed files in PR file tree

cleanjunc:
## Summary

Adds an "Only Changed" switch to the PR Files Changed sidebar. With the switch on (the default), the file tree shows only files that the pull request modifies. With the switch off, the tree expands to the full repository and unchanged files appear dimmed, matching the previous behavior.

The switch routes the tree builder through an empty source list when on, since the existing overlay pass already creates nodes for changed paths. The synthesized fallback list and a small block of duplicated stripping logic in the diff renderers are removed as part of the cleanup.

## Related Issues

Closes #957 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

[show-only-changed.webm](https://github.com/user-attachments/assets/ebd589a5-26d5-400e-aa72-8da98691f6c2)

## Checklist

- [x] New components are modularized/separated where sensible
- [x] Uses predefined theme (e.g. no hardcoded colors)
- [x] Responsive/mobile checked
- [x] Tested against the test API
- [x] `npm run format` and `npm run lint:fix` have been run
- [x] `npm run build` passes
- [x] Screenshots included for any UI/visual changes

PR #962, opened 2026-05-06
feat: inline word diff on PR Files Changed modified lines

### LINKED ISSUE #961: PR Files Changed should highlight only the changed tokens on modified lines, not the whole line
## Problem

When a pull request changes a single token on a long line, the renderer in `src/components/prs/PRFilesChanged.tsx` paints the entire line red on the left and the entire line green on the right. Reviewers then have to compare both copies of the line side by side just to find the characters that actually changed.

## Context

The renderer already pairs the left and right halves of modified lines. The split view's row builder carries `{ type: 'modify'; left: Change | null; right: Change | null }` exactly for this case, so the pairing work is done.

## Proposal

Add an inline word diff on those paired rows so only the tokens that actually changed are highlighted. A small in-file LCS pass, or a pull of `diffWords` from the `diff` package, fits without restructuring the viewer. The Unified view benefits from the same change since it walks the same paired rows.

This is the single biggest readability win the diff viewer can ship. It pairs naturally with the hide whitespace toggle, lives entirely client side, and matches what reviewers expect from GitHub's own diff UI.

cleanjunc:
## Summary

Highlights only the changed tokens on paired modify rows instead of tinting the entire line. The light line tint stays as ambient context, and a darker token tint sits on top of the actual edits, matching GitHub's own diff UI.

The split view (wrap and no wrap) and the unified view share a single row builder and segment renderer, so behavior stays consistent across modes. Unpaired additions and deletions keep the existing whole line tint, since there is no counterpart line to compare against.

## Related Issues

Closes #961

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Implementation Notes

The diff viewer now runs on a single uniform data model. Each visible line is represented as a `LineSide` with `{ kind: 'normal' | 'add' | 'del'; ln: number; segs: WordDiffSeg[] }`, and a `DiffRow` is either a chunk header or a paired `{ left, right }`. Two small lookup tables (`LINE_BG`, `TOKEN_BG`) map `kind` to the line and token tints, which removed the per cell ternary chains the old code carried.

The token segments themselves come from a small LCS based `wordDiff` over a tokenization of words, whitespace runs, and punctuation. Lines whose token count exceeds 200 on either side fall back to a whole line highlight to keep the DP table from growing on minified files. Tints reuse `DIFF_COLORS` via `alpha(..., 0.4)` for the token layer over the existing `alpha(..., 0.15)` line layer, so no new theme values were introduced.

Both views consume the same `buildDiffRows` output. The split view renders left and right through a shared `SplitSidePair` component, so the four cell duplication in wrap mode collapses into two component calls. The unified view flattens each paired row through `expandUnified`, which collapses normal context lines into a single combined row and emits two rows for paired modifies.

Net result on the file is a 97 line reduction, no new runtime dependencies, and a clearer separation between the data layer (row building plus word diff) and the render layer (cells plus tints).

## Screenshots

Before: every character of the modified line is tinted.
before

After: only the changed tokens carry the darker tint, with the light line tint preserved for context.
after
after1

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/prs/PRFilesChanged.tsx
+++ b/src/components/prs/PRFilesChanged.tsx
@@ -34,8 +34,10 @@ import ViewColumnIcon from '@mui/icons-material/ViewColumn'; // Split
 import axios from 'axios';
 
 import parseDiff, {
+  type AddChange,
   type Change,
   type Chunk,
+  type DeleteChange,
   type File as DiffFile,
 } from 'parse-diff';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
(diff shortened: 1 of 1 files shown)


cleanjunc:
@anderdc Could you please review this PR? Thanks!

anderdc:
Fold into #958 — same surface (PR Files Changed). Add the inline word-diff changes as commits on #958's branch.

PR #762, opened 2026-04-23
refactor: unused or misleading exported symbols

andriypolanski:
## Summary

Tightens the public module surface by removing or privatizing symbols that were exported but unused outside their defining files, and by dropping a dead dashboard hook. Aligns with the hygiene issues described in `bug-report-unused-exports-public-api-surface.md` (overlapping themes with [#739](https://github.com/entrius/gittensor-ui/issues/739), but includes additional items such as `StatRow`, duplicate named/default exports, and the `repositories` barrel re-export).

## Related Issues

Closes #761 

## Type of Change

- [x] Bug fix (maintainability / misleading API surface)
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## What Changed

| Area | Change |
|------|--------|
| `src/utils/ExplorerUtils.ts` | `SCORING_WINDOW_DAYS` is no longer exported; remains an internal constant for `isOutsideScoringWindow`. |
| `src/api/DashboardApi.ts` | `useDashboardQuery` is module-private. **`useCommitLog` removed** (zero in-repo callers; `useInfiniteCommitLog` unchanged). |
| `src/components/miners/TrustBadge.tsx` | `getRiskAssessment` is module-private. |
| `src/components/miners/index.ts` | Stop re-exporting `getRiskAssessment`. |
| `src/components/leaderboard/ActivitySidebarCards.tsx` | `StatRow` / `StatRowProps` are file-private. |
| `src/components/leaderboard/index.ts` | Export only `ActivitySidebarCards` (drop `StatRow` from barrel). |
| `src/components/repositories/index.ts` | Remove unused `resolveRelativeUrl` re-export (callers use `./MarkdownRenderers`). |
| `src/pages/search/SearchResultsCard.tsx` | Single export path: `const` + `export default` only. |
| `src/pages/dashboard/useDashboardData.ts` | Single export path: `const` + `export default` only. |

## Behavior / Risk

- **No UI or runtime behavior changes** intended for screens users interact with.
- **Breaking risk (external only):** Any code **outside this repo** that imported `useCommitLog` from the app’s API module would need to migrate (in-repo grep found no usages).

## Checklist

- [x] Scoped to dead / redundant exports and one unused hook
- [x] `npm run lint` passes
- [x] `npm run build` passes
- [ ] `npm run format` / `npm run lint:fix` run (if required by team policy)
- [ ] Screenshots N/A confirmed with reviewers

PR #979, opened 2026-05-07
fix(bounties): consolidate issues queries and stop loading flash

### LINKED ISSUE #978: [Bug] Bounties page fires three useIssues queries and shows partial data while two of them are still loading
`src/pages/IssuesPage.tsx` opens with three separate `useIssues` calls, then merges them by hand:

```tsx
const activeIssuesQuery = useIssues('active');
const registeredIssuesQuery = useIssues('registered');
const historyIssuesQuery = useIssues('completed,cancelled');

const allIssues = useMemo(() => {
  const seen = new Set<number>();
  const result = [];
  for (const issue of [
    ...(activeIssuesQuery.data || []),
    ...(registeredIssuesQuery.data || []),
    ...(historyIssuesQuery.data || []),
  ]) {
    if (!seen.has(issue.id)) {
      seen.add(issue.id);
      result.push(issue);
    }
  }
  return result;
}, [
  activeIssuesQuery.data,
  registeredIssuesQuery.data,
  historyIssuesQuery.data,
]);

const isLoading =
  activeIssuesQuery.isLoading &&
  registeredIssuesQuery.isLoading &&
  historyIssuesQuery.isLoading;
```

Three real problems with this shape:

**1. Three network round trips when one would do.**

The third call already passes a comma-separated value (`'completed,cancelled'`) which the API accepts. `src/api/IssuesApi.ts:16` shows `useIssues(status?, repository?)` simply forwards `status` as a query param, and three other call sites (`MinerRepositoriesTable.tsx:77`, `useDashboardData.ts:44`, `WatchlistPage.tsx:2301`) already pass no status and consume the unfiltered list. A single `useIssues('active,registered,completed,cancelled')`, or `useIssues()` if the backend default already covers all four buckets, replaces the fan-out with one request and one cache entry.

**2. The `isLoading` flag uses `&&`, which triggers a partial-data flash.**

`isLoading` is `true` only while *every* one of the three queries is still loading. As soon as the first one resolves, `isLoading` drops to `false` and the skeleton in `IssuesList` is replaced with whatever single bucket happened to land first. The other two finish moments later and rows pop in. The user sees the page repaint twice on a cold cache. The correct logic for "still loading" is `||` (any one is still loading), not `&&`.

**3. The dedupe keeps the first-seen copy, which can be the stale one.**

The `for` loop adds the issue to `result` only `if (!seen.has(issue.id))`, so when the same issue appears in two responses (which happens around status transitions, since `active` can briefly contain an issue that the backend has already moved to `registered`), the row picks up the data from whichever query returned first. There is no merge by `updatedAt`. The newer payload from `registered` is dropped on the floor.

Suggested fix:

* Replace the three calls with a single `useIssues()` (or `useIssues('active,registered,completed,cancelled')` for an explicit filter) and let `IssuesList` slice the response by tab.
* Drop the manual `useMemo` dedupe; one API response cannot contain duplicates.
* Drop the bespoke `isLoading` calculation and use the single query's `isLoading` directly.

Net effect: one request instead of three, a clean skeleton-then-data transition with no flash, and the bounty rows always reflect the most recent data the API returned.

cleanjunc:
## Summary

The Bounties page used to fire three separate `useIssues` requests (`active`, `registered`, `completed,cancelled`) and stitch the responses together. Two problems came out of that shape:

* `isLoading` was AND-ed across the three queries, so the skeleton vanished as soon as the first response landed and the remaining buckets popped in afterward. On a cold cache the page repainted twice.
* The manual dedupe in `useMemo` kept the first-seen copy of an issue. Around status transitions the same issue can briefly appear in two responses, so a stale row could win over a newer one.

This PR collapses the three calls into a single `useIssues()` and removes the bespoke merge. `IssuesList` already slices the response by tab (`active` / `registered` / `completed | cancelled`), so no consumer changes are needed. The unfiltered cache key is the same one already populated by `WatchlistPage`, `useDashboardData`, and `MinerRepositoriesTable`, so navigation between those pages now lands on a warm cache.

Net effect: one request instead of three, a clean skeleton-to-data transition with no flash, and the rows always reflect the latest payload from the API.

## Related Issues

Fixes #978

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

Not applicable. No visual change beyond the eliminated skeleton flash on cold load.

## Checklist

- [x] `npm run build` passes


cleanjunc:
@anderdc @e35ventura Could you please review this PR? Thanks!

anderdc:
Fix conflicts.

cleanjunc:
Hi @anderdc
I've fixed the conflicts. But `src/components/leaderboard/ActivitySidebarCards.tsx` is not related to this PR, the bug is from the just merged PR #762.
Could you please review again?
Thank you!

anderdc (approved):

PR #1005, opened 2026-05-09
fix(avatars): robust avatar resolution for miners using numeric IDs

watcharaponthod-code:
### Summary

This PR fixes a bug where avatars fail to load in the leaderboard and miner lists when a miner's author name (username) is missing and the system falls back to their numeric `githubId`. 

Previously, the system would attempt to fetch the avatar from `github.com/{id}.png`, which returns a 404 for numeric IDs. I have refactored the avatar utility to automatically detect numeric identifiers and use the correct `avatars.githubusercontent.com/u/{id}` endpoint.

### Changes

- **Utility Refactor (`src/utils/avatar.ts`):** Added `getGithubAvatarSrc`, a robust resolver that handles both usernames and numeric IDs with auto-detection logic.
- **Global Update (`src/utils/ExplorerUtils.ts`):** Re-exported the robust utility to ensure all callers benefit from the fix.
- **Leaderboard Fix (`LeaderboardSidebar.tsx`, `MinerCard.tsx`, `MinersList.tsx`):** Updated call sites to pass both `author` and `githubId`, providing the resolver with the best available data.

### Verification

- Verified that numeric strings (e.g., "240883001") now resolve to `https://avatars.githubusercontent.com/u/240883001` instead of `https://github.com/240883001.png`.
- Verified that alphabetic usernames still resolve to the standard `github.com/{user}.png` URL.
- Ensured no regressions in other components using the shared utility.

### DIFF
--- a/src/components/leaderboard/LeaderboardSidebar.tsx
+++ b/src/components/leaderboard/LeaderboardSidebar.tsx
@@ -274,7 +274,7 @@ const LeaderboardRow: React.FC<LeaderboardRowProps> = ({
         }}
       >
         <Avatar
-          src={getGithubAvatarSrc(miner.author || miner.githubId)}
+          src={getGithubAvatarSrc(miner.author, miner.githubId)}
           alt={miner.author || miner.githubId}
           sx={{ width: 20, height: 20 }}
         />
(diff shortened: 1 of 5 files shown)


ventura-oss:
⚠️ **Skipped during review** — No CI checks have run on this PR. CI requires maintainer approval for first-time contributors and won't auto-run regardless of new pushes. A maintainer will approve the workflow run when ready to review.

PR #1032, opened 2026-05-11
Fix sortable table header layout shift

### LINKED ISSUE #995: bug: table header layout shifts when sorting columns
## Description

Table column headers change layout/width when a sort is applied. Clicking a
sortable column header causes the header row to reflow — columns shift position
or resize — instead of staying stable.

## Steps to Reproduce

1. Navigate to any page with a sortable table (e.g. miner details → Pull Requests, Repositories, or Issues tab)
2. Click a column header to sort
3. Observe the header row layout shifting

## Expected Behaviour

Column widths and header layout remain fixed regardless of active sort state.
Only the sort indicator (arrow icon) should change.

## Actual Behaviour

Header cells reflow/resize when a sort column is selected, causing visual
instability in the table layout.

## Affected Components

- Any `DataTable` instance that uses `sortKey` on columns
- Observed on miner details page tables

## Possible Cause

The sort indicator icon appearing/disappearing on click may be pushing cell
content and causing width recalculation. Column widths may not be explicitly
fixed, allowing the header to reflow when the icon is injected into the DOM.

lby910722-ship-it:
## Summary
- reserve a fixed icon track for sortable `DataTable` headers
- keep the sort icon mounted for inactive headers so sorting only changes icon state, not header width

Closes #995

## Testing
- `npm run build`
- `npm run lint`
- `npm test`
- `npx prettier --check src/components/common/DataTable.tsx`
- `git diff --check`

## Notes
- Tried local visual checks at `/watchlist` and `/bounties`, but this fresh local run currently hits existing data-shape errors (`miners.map is not a function`, `allIssues.filter is not a function`) before the affected tables render.

### DIFF
--- a/src/components/common/DataTable.tsx
+++ b/src/components/common/DataTable.tsx
@@ -105,6 +105,22 @@ const tableSx = {
   width: '100%',
 } as const;
 
+const sortLabelSx: SxProps<Theme> = {
+  display: 'inline-grid',
+  gridTemplateColumns: 'minmax(0, auto) 16px',
+  alignItems: 'center',
+  columnGap: 0.5,
+  maxWidth: '100%',
+  verticalAlign: 'middle',
+  '& .MuiTableSortLabel-icon': {
+    justifySelf: 'center',
+    m: 0,
+    width: 16,
+    height: 16,
+    flexShrink: 0,
+  },
+};
+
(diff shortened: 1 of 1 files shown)


anderdc:

PR #1041, opened 2026-05-11
refactor(repositories): migrate leaderboard URL state onto useDataTableParams hook

### LINKED ISSUE #1040: Consolidate TopRepositoriesTable URL state onto the shared useDataTableParams hook
## Summary

The two leaderboard tables in `src/components/leaderboard/` are conceptually the same: a sortable, filterable, paginated table with URL-backed state so visitors can bookmark and share filtered views. They diverge sharply in implementation.

`TopMinersTable.tsx` uses the shared `useDataTableParams` hook from `src/hooks/useDataTableParams.ts`. The hook handles parsing each param from the URL on render, computing the next sort direction when a column is re-clicked, dropping default values from the URL, and writing back via `setSearchParams`. About 20 lines at the call site.

`TopRepositoriesTable.tsx` does the same job by hand:

* Seven `useState` slots initialized from the URL on first render.
* An `isInitialMount` ref to guard the first effect run.
* An `eslint-disable react-hooks/exhaustive-deps` suppression on the search-reset effect.
* A `syncToUrl` callback that has to enumerate every key the table uses.
* Roughly 150 lines of plumbing for the same contract the hook covers in around 40.

## Drift between the two implementations

* `syncToUrl` calls `setSearchParams(params, { replace: true })`, which replaces the entire search string on every write. Any query parameter outside the table's own keys is dropped.
* The default-rows check `if (rows !== '10')` only matches the list-view default. In cards view the default is 12, so the URL keeps `rows=12` even when it matches the cards default.
* The `view=list` value is silently dropped from the URL, so deep-linked list-mode shares depend on the recipient's localStorage matching.

Any cross-cutting URL-state work has to be applied to both implementations rather than to the hook alone.

## Suggested change

Migrate `TopRepositoriesTable` onto `useDataTableParams`, matching the existing usage in `TopMinersTable`. The mapping is mechanical, the visible contract is preserved, and the two leaderboard tables converge on one path forward.

blinkeye-lcm:
## Summary

`TopRepositoriesTable` now reads and writes its sort, filter, view-mode, and pagination state through the shared `useDataTableParams` hook, matching how `TopMinersTable` already works.

The hand-rolled `syncToUrl` callback, the `isInitialMount` ref, the page-reset `useEffect` with its `eslint-disable react-hooks/exhaustive-deps` suppression, and the URL-tracking `useState` slots all go away. A single `useDataTableParams` call plus a typed `filtersConfig` (search, status, view) covers the same surface in roughly a third of the code.

Display behavior is preserved for every interaction (sort, filter, search, pagination, view toggle). Two URL-representation changes are intentional and converge with `TopMinersTable`:

* `view=cards` or `view=list` is always written after a toggle. The comment in `TopMinersTable.tsx` documents why dropping the param on the default value can leave the UI stuck on a stale localStorage value.
* Trailing whitespace in the search box is trimmed at serialize time so the URL stays clean.

The cards-vs-list page-size asymmetry (10/25/50 vs 12/24/48) is handled by clamping the hook's raw value through the existing `clampRowsForRepositoriesView` helper for display, and by reflowing the URL on view toggle when the current value is invalid for the new view.

No user-visible bug is being fixed. The latent benefits include:

* Future in-app navigation to `/repositories?...` from inside the SPA will be picked up by the table without a re-mount.
* Other components can add query parameters to `/repositories` without the table clobbering them on its next write.
* Cross-cutting URL-state improvements only need to land in `useDataTableParams`.

## Related Issues

Closes #1040 

## Type of Change

* [ ] Bug fix
* [ ] New feature
* [x] Refactor
* [ ] Documentation
* [ ] Other (describe below)

## Screenshots

No visible UI change. Verified locally that sort, status filter, search, pagination, and view toggle all behave as before. URL representation matches `TopMinersTable` for the two intentional convergence points noted above.

## Checklist

* [ ] New components are modularized/separated where sensible
* [x] `npm run build` passes
* [ ] Screenshots included for any UI/visual changes


anderdc:
fix conflicts

blinkeye-lcm:
Hi @anderdc , I've fixed the conflicts. Could you review it again please?

anderdc (approved):

Issue #685, opened 2026-04-21
[BUG] Dashboard Live Activity infinite scroll and repo table bar chart are wrong


corevibe555:
## Description
[Note] Please check  Issue 1, Issue 2.

**Issue 1.  Dashboard live activity infinite scroll** 
Dashboard Live Activity Infinite scroll doesn't load properly after some scroll events. 
This happens often when I scroll down a bit faster pace  but not super fast.

## Steps to Reproduce

1. Visit https://gittensor.io/dashboard
2. Scroll down Live Activity section
3. Keep scrolling down the PRs list, a bit faster.
4. stop scrolling, you will see the loading indicator loads indefinitely and doesn't load any data.
See network tab, you will see  next page fetch API request sent infinitely though I stopped scrolling.

## Expected Behavior

Data should load without issue when I stopped scrolling.

## Actual Behavior

Data doesn't load properly in this scenario.
[Note] This happens when I didn't open Network tab as well. Confirmed.
## Additional Context

https://github.com/user-attachments/assets/79b0d8e4-903b-46aa-bc1d-bdfa78552f84

https://github.com/user-attachments/assets/ecf12fc7-36ed-4b33-9cf4-dc62a08b0bf3

Please review `getNextPageParam`, `return allPages.length + 1;`
Might be good to use `return lastPageParam + 1;`

Please validate on your own.

**Issue 2. Repo table chart**
## Steps to Reproduce

1. Visit https://gittensor.io/repositories?status=active
2. Click on "Active 214" tab on repo list table.
3. Turn on "Log Scale" switch.
4. Scroll down to Repository list table and click on "Weight" column for sorting
5. Observe when weight is ascending order, the repo weight bar chart shows identical height bars.(weight 1 and weight 0.14 show the same bar height)
6. Observe when weight is in descending order, the repo weight bar chat shows nothing.

## Expected Behavior
Chart should show valid info

## Actual Behavior
Chart shows wrong info

https://github.com/user-attachments/assets/a050ec8d-fdd1-4090-b516-0c6e09aa26ad

## Environment

- Browser: Chrome
- OS: Windows

anderdc:
Resolved by merged PR #1056, which stabilized the Live Activity infinite scroll. The second sub-issue (repo table bar chart heights) was investigated in that PR and could not be reproduced. Closing as completed — please open a fresh issue with repro steps if the chart problem resurfaces.

PR #1056, opened 2026-05-12
fix: stabilize Live Activity infinite scroll pagination

nickmopen:
## Summary

Fixes the Live Activity infinite scroll on the Dashboard that would enter an endless loading loop when scrolling at a moderate pace.

> **Note:** This PR addresses **Issue 1** (infinite scroll) from #685. **Issue 2** (repo table chart showing incorrect bar heights / empty bars when sorting by Weight with Log Scale enabled) was investigated but could **not be reproduced** — the chart behaved correctly under all tested conditions (ascending, descending, log scale on/off). No changes were made for Issue 2.

Two root causes addressed for Issue 1:

- **`DashboardApi`**: `getNextPageParam` was using `allPages.length + 1` to derive the next page number. During a background `refetchInterval` refetch, `allPages` is rebuilt page-by-page and can be shorter than expected when `getNextPageParam` is evaluated, causing a stale/incorrect page number to be requested. Changed to `lastPageParam + 1` which is always the canonical value.

- **`LiveCommitLog`**: The `IntersectionObserver` had `isFetchingNextPage` and `visibleEntries.length` in its dependency array. Every time a page loaded, the effect would teardown and recreate the observer — which would immediately fire if the sentinel element was still in the viewport, bypassing the in-flight guard and triggering another `fetchNextPage()`. This created the infinite request loop. Replaced with a debounced scroll handler (120ms) that checks a fixed pixel distance from the bottom of the container, with a `fetchInFlightRef` flag to prevent concurrent fetches. Removes the sentinel DOM element entirely.

## Related Issues

See #685 (Issue 1 only — Issue 2 not reproducible, no changes made)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## before
https://github.com/user-attachments/assets/4b8f68a7-d1dd-45c9-bcab-43d99a6350ce

## after
https://github.com/user-attachments/assets/4fef8ee9-918c-4727-9aeb-b5647bf693b6

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/api/DashboardApi.ts
+++ b/src/api/DashboardApi.ts
@@ -51,13 +51,17 @@ export const useInfiniteCommitLog = (options?: {
       });
       return data;
     },
-    getNextPageParam: (lastPage: CommitLog[], allPages: CommitLog[][]) => {
+    getNextPageParam: (
+      lastPage: CommitLog[],
+      _allPages: CommitLog[][],
+      lastPageParam: number,
+    ) => {
       // If the last page has fewer items than the limit, we've reached the end
       if (lastPage.length < limit) {
         return undefined;
       }
-      // Otherwise, return the next page number
-      return allPages.length + 1;
+      // Use lastPageParam to avoid stale allPages count during background refetches
+      return lastPageParam + 1;
     },
     initialPageParam: 1,
     refetchInterval: options?.refetchInterval,
--- a/src/pages/dashboard/views/LiveCommitLog.tsx
+++ b/src/pages/dashboard/views/LiveCommitLog.tsx
@@ -1,4 +1,10 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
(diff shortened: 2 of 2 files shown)


anderdc (changes requested):
Three things to address before merge:

- src/pages/dashboard/views/LiveCommitLog.tsx: the scrollDebounceRef timeout is never cleared on unmount. Add a cleanup effect that clears the pending timeout.
- LiveCommitLog.tsx: the scroll handler only fires on a real scroll event, so if the first page doesn't fill the container there's nothing to trigger fetchNextPage and more pages can't load. Handle the not-scrollable case — the IntersectionObserver this replaces covered it automatically.
- PR body: "Fixes #685" will auto-close the issue on merge while Issue 2 (repo bar chart) is still unaddressed. Reference #685 without the closing keyword so it stays open for Issue 2.

anderdc (approved):

PR #1060, opened 2026-05-12
fix(discoveries): sort Credibility column by issue credibility

omipheo:
## Summary

On the Discoveries leaderboard the **Credibility** column never produces
a real ordering: each `MinerCard` already displays `issueCredibility`
(selected via the `isDiscoveries` branch), but the comparator in
`TopMinersTable` sorts by `miner.credibility` instead. `DiscoveriesPage`'s
local stats mapping never populates `credibility`, so every row ties
at `0` and the sort has no visible effect — confirmed by the reporter's
screenshots in #787 (issue 4).

This PR passes the `LeaderboardVariant` into `compareMiners` and reads
`issueCredibility` on the `discoveries` variant so the sort key matches
the displayed value. OSS and Watchlist variants keep reading
`credibility` unchanged.

Adds 4 new vitest cases in `src/tests/topMinersSort.test.ts` covering
both variants and the missing-data fallback, mirroring the existing
`src/tests/watchlistSort.test.ts` pattern.

## What changed

- `src/components/leaderboard/TopMinersTable.tsx`: `compareMiners`
  exported and accepts an optional `variant` parameter (defaults to
  `'oss'` so any future caller stays at parity with current behavior).
  The `case 'credibility':` branch now reads `issueCredibility` when
  `variant === 'discoveries'`. The `useMemo` deps for `rankedMiners`
  pick up `variant`.
- `src/tests/topMinersSort.test.ts` (new): 4 unit tests covering OSS
  variant, Discoveries variant, missing-data fallback, and the default
  parameter.

## Verification

- `npm run test` — 79 tests, 5 files, all passing.
- `npm run build` — TypeScript + Vite, no errors.
- `npm run lint` — no warnings.
- `npm run format:check` — clean.

## Scope

Closes part of #787 — specifically issue 4 ("Sort by Credibility not
working on Discoveries page"). The other three bugs in #787
(Open/Closed filter case, duplicate issue rows, mobile table layout)
are intentionally left to a focused follow-up rather than bundled
here. Bug 1 in particular doesn't reproduce against the current
`RepositoryIssuesTable.tsx`, which already filters on `!issue.closedAt`
rather than a `state === 'CLOSED'` string.

### DIFF
--- a/src/components/leaderboard/TopMinersTable.tsx
+++ b/src/components/leaderboard/TopMinersTable.tsx
@@ -172,11 +172,12 @@ const passesSingleProgramEligibility = (
   filter === 'all' ||
   (filter === 'eligible' ? Boolean(miner.isEligible) : !miner.isEligible);
 
-const compareMiners = (
+export const compareMiners = (
   a: MinerStats,
   b: MinerStats,
   option: SortOption,
   isWatched: (key: string) => boolean,
+  variant: LeaderboardVariant = 'oss',
 ): number => {
   switch (option) {
     case 'totalScore':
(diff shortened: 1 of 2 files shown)


anderdc:

Issue #1083, opened 2026-05-13
[Bug] Typing in list search is sluggish when tables have large datasets


bcExpt1123:
## Description

Searching in list/table views (especially when there are **many rows** loaded) causes **noticeable lag or jank while typing** in the search field. The UI can feel like it is struggling to keep up with each character, which hurts **perceived performance and usability**.

The problem is **most obvious** on **Watchlist → PRs** (`WatchlistPage`), where users may have a **large number of watched PRs** and the table does substantial work on each interaction. Similar list-heavy screens can show the same behavior to a **lesser degree** when datasets are smaller, so the issue may go unnoticed even though the underlying pain exists.

## Steps to Reproduce

1. Open the app and go to **Watchlist**.
2. Open the **PRs** tab and ensure you have a **large set of watched PRs** loaded (the more rows, the clearer the effect).
3. Click into the **search** field on filters for that tab.
4. Type a search string **continuously** at normal speed (e.g. several characters per second) and observe the field and table.

*(Optional: repeat on other list views with many rows—issues list, repositories leaderboard, miner repositories table—to compare severity.)*

## Expected Behavior

- The search field should **stay responsive** while typing (characters appear immediately, cursor movement feels normal).
- The page should **remain usable** while refining a query: scrolling and other controls should not feel stuck.
- Filtering/narrowing the list may update **shortly after** the user pauses or finishes typing, without making typing itself feel heavy.

## Actual Behavior

- While typing in search, the experience can feel **sluggish or stuttery**, especially with many PRs visible.
- The problem may be **subtle or absent** on tabs/pages with **fewer rows**, so it is easy to assume “everything is fine” globally even when large lists misbehave.

## Additional Context

**Primary repro:** Watchlist **PRs** tab — `src/pages/WatchlistPage.tsx` (large `DataTable` / filtered PR list).

**Other surfaces where the same class of issue can appear** (search over non-trivial client-side lists or dense UI):

| Area | File |
|------|------|
| Watchlist – Repositories | `src/pages/WatchlistPage.tsx` |
| Watchlist – Bounties | `src/pages/WatchlistPage.tsx` |
| Watchlist – Issues (watched miners) | `src/pages/WatchlistPage.tsx` |
| Issues page | `src/components/issues/IssuesList.tsx` |
| Repositories leaderboard | `src/components/leaderboard/TopRepositoriesTable.tsx` |
| Miner repositories table | `src/components/miners/MinerRepositoriesTable.tsx` |
[...]

PR #1087, opened 2026-05-13
Fix: Use deferred value for list search to prevent typing lag

Stevenn28:
## Description\n\nCloses #1083.\n\nThis PR fixes the noticeable typing lag in Watchlist search fields (especially on the PRs tab with many rows). The lag was caused by state updates triggering synchronous filtering and rendering on every keystroke.\n\nBy introducing `useDeferredValue` for the search query, the input field state (`searchQuery`) remains highly responsive for typing, while the expensive `useMemo` filtering blocks are deferred until React has idle time, maintaining a smooth user experience.\n\n## Changes\n- Imported `useDeferredValue` in `WatchlistPage.tsx`\n- Added `const deferredSearchQuery = useDeferredValue(searchQuery)` to all major list components (`ReposList`, `BountiesList`, `PRsList`, `IssuesList`).\n- Updated the `filter` functions and dependency arrays to use `deferredSearchQuery` instead of the immediate `searchQuery`.

### DIFF
--- a/src/pages/WatchlistPage.tsx
+++ b/src/pages/WatchlistPage.tsx
@@ -4,6 +4,7 @@ import React, {
   useMemo,
   useState,
   useRef,
+  useDeferredValue,
 } from 'react';
 import {
   Avatar,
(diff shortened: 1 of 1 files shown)


anderdc:

PR #1115, opened 2026-05-13
fix(miner-details): make filter button counts respect search query

### LINKED ISSUE #1114: Filter button counts on Miner Details tabs don't update when typing in the search box
## Summary

On the Miner Details page, the row counts on the **All / Open / Merged / Closed** filter buttons (PR tab) and **All / Open / Solved / Closed** filter buttons (Open Issues tab in Issue Discovery mode) are computed from the full dataset and never react to the search input. Typing in the search box filters the table rows but leaves the button counts stale, so the user sees e.g. "Open (42)" even when the visible list has shrunk to one row.

## Where

- `src/components/miners/MinerPRsTable.tsx` — Pull Requests tab
- `src/components/miners/MinerOpenDiscoveryIssuesByRepo.tsx` — Open Issues tab (both the "Your open discovery issues" and "Other open discovery issues" sections; each has its own search box and counts)

## Steps to reproduce

1. Open any miner profile, e.g. `/miners/details?githubId=…`.
2. Switch to the **Pull Requests** tab. Note the counts in the four filter buttons.
3. Type a string into the search box that matches few rows.
4. The table shrinks, but the four button counts are unchanged.
5. Switch to **Issue Discovery → Open Issues**. Same behavior in both sections.

## Expected

Each button shows the number of rows that *would* be selected if you clicked it under the current **search + author** scope. The currently-active status filter itself is excluded from the scope so the inactive buttons keep meaningful counts.

## Actual

Counts reflect the raw, unfiltered list and never change with the search input.

## Root cause

The `useMemo` that builds the counts depends only on the source data (`prs` / `mineIssues` / `otherIssues`) and skips the search query and author chip.

`MinerPRsTable.tsx:203`:
```ts
const statusCounts = useMemo(() => {
  if (!prs) return { all: 0, open: 0, merged: 0, closed: 0 };
  return getPrStatusCounts(prs);   // ignores searchQuery + selectedAuthor
}, [prs]);

plind-junior:
## Summary

Filter button counts on the Miner Details page now update live as the user types in the search box. Previously the All/Open/Merged/Closed (PR tab) and All/Open/Solved/Closed (Open Issues tab) counts were computed from the full unfiltered dataset, so users saw stale numbers like "Open (42)" even when the visible list had shrunk to one row.

Fixes the bug described in the linked issue.
Fixes #1114 

### `src/components/miners/MinerPRsTable.tsx`

Scope the count computation to the same `search + author` filters that drive the visible table, with `statusFilter: 'all'` so the active button doesn't zero out the others.

​```ts
const statusCounts = useMemo(() => {
  if (!prs) return { all: 0, open: 0, merged: 0, closed: 0 };
  const scope = filterPrs(prs, {
    author: selectedAuthor,
    includeNumber: true,
    searchQuery,
    statusFilter: 'all',
  });
  return getPrStatusCounts(scope);
}, [prs, selectedAuthor, searchQuery]);
​```

### `src/components/miners/MinerOpenDiscoveryIssuesByRepo.tsx`

- Extracted the issue-search filter from `applyIssueFilter` into a reusable `applyIssueSearch` helper.
- `mineCounts` and `otherCounts` now run through `applyIssueSearch` with each section's own search query (`mineSearch` / `otherSearch`), so the two sections' filter buttons update independently.

## How counts behave now

| Scope used for counts | Search | Author chip | Active status filter |
|---|---|---|---|
| Active | ✓ | ✓ (PR tab only) | ✗ (excluded so inactive buttons stay meaningful) |

Each button shows the number of rows that *would* be selected if you clicked it under the current search/author scope.

## Test plan

- [ ] Open `/miners/details?githubId=…` → **Pull Requests** tab.
- [ ] Note counts in the four filter buttons.
- [ ] Type a query that matches few rows — counts shrink with the table.
- [ ] Clear the query — counts return to full dataset.
- [ ] Click an author chip — counts reflect just that author.
- [ ] Switch to **Issue Discovery → Open Issues**.
- [ ] Type in the **"Your open discovery issues"** search — its 4 buttons update, the "Other" section's buttons are unaffected.
- [ ] Type in the **"Other open discovery issues"** search — vice versa.
- [ ] `npm run type-check` passes.

## Notes

- Active status filter is intentionally excluded from the count scope. Including it would zero out every other button when one is active, which defeats the purpose of having a filter bar.
- No new dependencies; reuses existing `filterPrs` and refactors a small inline filter into `applyIssueSearch`.

https://github.com/user-attachments/assets/2948c689-f30c-46d4-93a7-3dd54966193c


anderdc (changes requested):
In MinerScoreBreakdown.tsx the new count and the visible table use different search matchers. The count (statusCounts, ~L1040) goes through filterPrs, which matches a PR number only as String(number). The visible table (filteredPrs, L1062-1074) additionally matches `#${number}`. So searching #42 shows the row but the count buttons exclude it — the stale-count bug this PR fixes, reintroduced for #-prefixed number searches. Route both the count and the visible table through one shared matcher, the way the other three files in this PR reuse the table's filter function for the count.

anderdc (approved):

PR #463, opened 2026-04-17
fix: inconsistent scrollbar usage across 10+ files and components

dataCenter430:
## Summary

- Replaced all custom inline `::-webkit-scrollbar` styles with the shared `scrollbarSx` constant from `theme.ts`
- Applied `scrollbarSx` to scrollable containers that had no scrollbar styling at all
- Ensures consistent scrollbar appearance (size, color, hover state) across every page and component

### Replaced custom inline styles (7 files)
- `src/pages/search/SearchResultsTable.tsx`
- `src/pages/dashboard/views/LiveCommitLog.tsx`
- `src/pages/dashboard/DashboardPage.tsx`
- `src/components/prs/PRFilesChanged.tsx`
- `src/components/miners/MinerRepositoriesTable.styles.ts`
- `src/components/miners/MinerPRsTable.tsx`
- `src/components/layout/GlobalSearchBar.tsx`

### Added missing scrollbar styling (10 files)
- `src/components/layout/AppLayout.tsx` — main page scroll container
- `src/components/leaderboard/LeaderboardSidebar.tsx`
- `src/components/repositories/FileExplorer.tsx`
- `src/components/ContributionHeatmap.tsx`
- `src/components/onboard/GettingStarted.tsx`
- `src/components/issues/IssuesList.tsx`
- `src/components/issues/IssueSubmissionsTable.tsx`
- `src/components/prs/PRComments.tsx`
- `src/components/issues/IssueConversation.tsx`
- `src/components/ErrorFallback.tsx`
- `src/pages/NotFoundPage.tsx`

### Screenshots

https://github.com/user-attachments/assets/ada29693-7fb9-4976-b617-bf0e92055d55

PR #286, opened 2026-04-15
refactor: replace inline scrollbar styles with shared scrollbarSx

wdeveloper16:
## Summary

- Replaces raw `&::-webkit-scrollbar` CSS blocks in 4 files with the shared `scrollbarSx` constant from `src/theme.ts`
- Follows the same pattern established by #249, which unified scrollbar styles across the rest of the codebase
- No functional changes — purely cosmetic/consistency cleanup

Files updated:
- `src/pages/search/SearchResultsTable.tsx`
- `src/components/layout/GlobalSearchBar.tsx`
- `src/components/miners/MinerPRsTable.tsx`
- `src/components/dashboard/LiveCommitLog.tsx`

Net: 10 insertions, 40 deletions across 4 files.

## Test plan

- [x] `npm run format:check` — passes
- [x] `npm run lint` — passes (0 warnings)
- [x] `npm run build` — clean build, no TS errors
- [x] Scrollbars render correctly in Search results, Global search dropdown, Miner PRs table, and Live commit log

PR #249, opened 2026-04-14
fix: unify scrollbar styles across code viewer and app

dataCenter430:
## Summary
- The code viewer's syntax-highlighted code area was showing the browser-default scrollbar because `react-syntax-highlighter` renders its own `<pre>` element that controls scrolling — custom scrollbar styles on the outer wrapper had no effect. Fixed by targeting `& pre` directly and setting the outer Box to `overflow: hidden`.
- The markdown view inside the code viewer was also missing custom scrollbar styles.
- Extracted the repeated scrollbar style block into a shared `scrollbarSx` constant exported from `theme.ts` and replaced 13 duplicate inline definitions across 12 files with `...scrollbarSx`.

## Changes
- `theme.ts` — added exported `scrollbarSx` constant (`8px` width/height, transparent track, semi-transparent thumb with hover)
- `CodeViewer.tsx` — scoped scrollbar styles to the inner `<pre>` element for the syntax view; added matching styles to the markdown view's scroll container
- 11 additional files — replaced inline scrollbar blocks with `...scrollbarSx`

## Test plan
- [x] Open a repository's Code tab and navigate into a file — scrollbar should be thin and dark-themed
- [x] Verify the same for a markdown file
- [x] Confirm the scrollbar matches the one on the Discoveries and Issues pages

## Screenshot or Video 

Before:
image

image

After:

https://github.com/user-attachments/assets/d467240a-115e-435a-a181-802bd68bde90

image

PR #1116, opened 2026-05-13
fix(scrollbar): Fix weird scrollbar issue on onboard page

plind-junior:
## Summary

On the Bounties page (and intermittently elsewhere), the browser's default light/system scrollbar would appear when scrolling — most reliably reproducible by applying a filter and then scrolling up and down. The dark theme scrollbar from `scrollbarSx` was only being applied to specific containers, so any other scroll context (the body itself, or a transient overflow context after a filter resize) fell back to the OS default.

This PR adds a global `::-webkit-scrollbar` + Firefox `scrollbar-color` baseline in `src/index.css` so every scrollable element on the page picks up the dark theme automatically.

## Cause

`scrollbarSx` ([src/theme.ts:82](../blob/test/src/theme.ts#L82)) is an MUI `sx` snippet that styles the scrollbar of the element it's attached to. Components have to opt in. Today only a handful of containers do — `<main>` in [AppLayout.tsx:154](../blob/test/src/components/layout/AppLayout.tsx#L154), DataTable's `TableContainer`, the global search results dropdown, etc.

When filtering on `/bounties` causes the result count (and therefore page height) to shrink, the scroll context can briefly move from `<main>` to the body for a frame, and the body's scrollbar isn't styled — producing the visible "default scrollbar appearing" flash.

## Related prior work

This is complementary to past per-container fixes — none of which targeted the global stylesheet:
- #249 — extracted `scrollbarSx` and applied it to ~12 components
- #463 — added `scrollbarSx` to AppLayout's `<main>` and 16 other containers
- #228 (closed) — attempted unified scroll container approach
- #286 (closed) — refactor follow-up to #249

## Test plan

- [ ] Open `/bounties`, apply a filter (e.g. "Available"), scroll up and down — no white/native scrollbar appears.
- [ ] Repeat on Watchlist, Repositories, Top Miners — scrollbar styling is consistent.
- [ ] Sanity: existing scrollbars on `<main>`, DataTable, and the global search results dropdown are visually unchanged.
- [ ] Firefox: scrollbar shows the thin dark variant via `scrollbar-color`.

## Out of scope

Not changing any per-container `scrollbarSx` usage — those keep working and are kept for explicit intent on specific scroll regions.

Before

https://github.com/user-attachments/assets/d8ad0851-95bc-4e77-aa75-6b370ae97e11

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -11,14 +11,15 @@
 }
 
 html,
-body {
+body,
+#root {
   margin: 0;
   padding: 0;
-  /* AppLayout's <main> is the only scroll container. Locking html/body
-   * prevents the native body scrollbar from showing alongside main's
-   * styled scrollbar. */
-  height: 100%;
-  overflow: hidden;
+  /* AppLayout's <main> is the only scroll container. Locking html/body/#root
+   * prevents the native browser scrollbar from showing alongside main's
+   * styled scrollbar. !important guards against any later CSS-in-JS resets. */
+  height: 100% !important;
+  overflow: hidden !important;
 }
 
 body {
--- a/src/index.css
+++ b/src/index.css
@@ -10,10 +10,18 @@
   color: #ffffff;
 }
 
+html,
 body {
   margin: 0;
   padding: 0;
-  overflow-x: hidden;
+  /* AppLayout's <main> is the only scroll container. Locking html/body
+   * prevents the native body scrollbar from showing alongside main's
+   * styled scrollbar. */
+  height: 100%;
+  overflow: hidden;
+}
+
+body {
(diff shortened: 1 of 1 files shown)


anderdc:
AppLayout's root Box is already height: 100dvh with overflow: hidden, and every route renders inside it, so the document/body can't grow a vertical scrollbar — this overflow: hidden !important lock removes a scrollbar that never appears. The actual scrollbars on the onboard page (<main> and GettingStarted's inner box) are both already styled via scrollbarSx and aren't touched. The description also doesn't match the diff — it claims a ::-webkit-scrollbar baseline that the diff never adds. Closing.

PR #1130, opened 2026-05-14
fix: `validateDOMNesting(...)` while keeping middle-click supports and _blank supports with `a` 

### LINKED ISSUE #1127: [BUG] `validateDOMNesting(...): <a> cannot appear as a child of <tbody>`
## Description

When rendering a DataTable component, React throws the following warning:
```
validateDOMNesting(...): <a> cannot appear as a child of <tbody>.
```

This indicates that an `<a>` (anchor) element is being rendered directly inside a `<tbody>`, which violates HTML DOM nesting rules.

## Steps to Reproduce

1. Open the Leaderboard page or any page where the DataTable is rendered.
2. Launch browser Developer Tools (e.g., right-click → Inspect).
3. Go to the Console tab.
4. Notice the following error message displayed:

Image

## Expected Behavior

Anchor (`<a>`) elements should only appear inside table cells (`<td>` or `<th>`), not directly under `<tbody>`.

## Actual Behavior

The `DataTable` renders (or allows rendering of) an `<a>` element directly under `<tbody>`, causing React to emit a DOM nesting warning.

## Impact
- Violates HTML standards
- Produces React warnings in development
- May lead to accessibility issues
- Could break layout or behavior in stricter environments

## Additional Context

This appears when implementing clickable rows or embedding links in table content. The issue may stem from how the DataTable internally structures rows.

bcExpt1123:
## Summary

Fixes invalid table DOM and restores native link UX for `DataTable` rows when `getRowHref` is set.

Previously, linked rows used `TableRow` with `component="a"`, which produced an `<a>` as a direct child of `<tbody>` and triggered React’s `validateDOMNesting` warning. That pattern also could not show the real URL in the browser status bar the way a normal link does.

`LinkTableRow` now renders a standard `<tr>` and clones each `TableCell` child to inject a stretched `<a href>` (via `useLinkBehavior`, same as `LinkBox`) plus a `pointer-events` layer so hover and primary clicks hit a real anchor while nested controls (`button`, `input`, inner links with `role="button"`, etc.) stay interactive. Secondary cell anchors are `aria-hidden` / `tabIndex={-1}` so keyboard focus stays sensible; only the first cell’s anchor is in the tab order by default.

`DataTable`’s `getRowHref` JSDoc is updated to describe the valid DOM and the requirement that cells must not nest another `<a>` when row links are enabled.

## Related Issues

fixes #1127 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

Before:

image

After:

image

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/common/DataTable.tsx
+++ b/src/components/common/DataTable.tsx
@@ -56,10 +56,10 @@ export type DataTableProps<T, SortKey extends string = never> = {
   emptyState?: React.ReactNode;
   minWidth?: number | string;
   /**
-   * When set, each row renders as an `<a href>` (native new-tab / middle-click
-   * support). NOTE: because the row becomes `<a>`, cells must not contain
-   * nested `<a>` elements (invalid HTML). For tables with nested anchors,
-   * use `onRowClick` instead.
+   * When set, each row stays a valid `<tr>` and each cell gets a stretched
+   * `<a href>` so the browser status URL, middle-click, and link context menu
+   * behave like a normal anchor. Cells must not contain nested `<a>` elements.
+   * For tables with nested anchors, use `onRowClick` instead.
    */
   getRowHref?: (item: T) => string;
   linkState?: Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


anderdc:
Closing — not a change we're looking to take on.

PR #1139, opened 2026-05-14
fix(leaderboard): floor credibility in list view instead of rounding

plind-junior:
## Summary

The Credibility column in the miner leaderboard list view (`MinersList.tsx`) rounds half-up via `toFixed(0)`, which can over-state a miner's credibility. A miner with `credibility = 0.967` is displayed as **`97%`** even though they haven't crossed that threshold. This PR switches to integer truncation so values floor toward the displayed integer — `96.7%` now displays as **`96%`**.

## Why

`toFixed(0)` rounds half-up — values in `[0.965, 0.974]` all become `97%`. For credibility (which gates eligibility at 80%), displaying a higher value than the miner has actually earned is misleading. Truncation via `Math.floor` is the conservative choice: the displayed integer is always ≤ the true value, so the user never sees a percentage they don't yet have.

## Behavior

| Raw credibility | Before (`toFixed(0)`) | After (`Math.floor`) |
|---|---|---|
| 0.7949 | 79% | 79% |
| 0.7950 | 80% | 79% |
| 0.8499 | 85% | 84% |
| 0.8500 | 85% | 85% |
| 0.967  | 97% | 96% |
| 1.000  | 100% | 100% |

Sort behavior is unchanged: ordering is still done on the raw fraction (`(a.credibility ?? 0) - (b.credibility ?? 0)` in `TopMinersTable.tsx:197-198`), so two miners at `0.847` and `0.852` still resolve order correctly even though both render as `84%` and `85%` respectively.

## Test plan

- [ ] Open `/leaderboard` (or any view that mounts `MinersList`).
- [ ] Switch to list view.
- [ ] Find a miner whose card-view donut shows ≥ X.5% — confirm the list-view Credibility column now reads `X%` (truncated) instead of `X+1%`.
- [ ] Sort by Credibility ascending and descending — order is unchanged.

## Out of scope

- Card-view `MinerCard.tsx` uses the unrounded percent directly in a donut visualization — no rounding change needed there.
- The `Credibility` cell in `MinerScoreBreakdown.tsx` and any miner-stats KPI tiles are unchanged. If we want a globally consistent truncation policy, that's a follow-up.

Before
image

After
image

### DIFF
--- a/src/components/leaderboard/MinersList.tsx
+++ b/src/components/leaderboard/MinersList.tsx
@@ -160,7 +160,7 @@ export const MinersList: React.FC<MinersListProps> = ({
       sortKey: 'credibility',
       renderCell: (miner) => (
         <Typography sx={{ ...cellTypographySx, color: 'text.primary' }}>
-          {((miner.credibility ?? 0) * 100).toFixed(0)}%
+          {Math.floor((miner.credibility ?? 0) * 100)}%
         </Typography>
       ),
     },


anderdc:
`toFixed(0)` is the credibility-display convention across the codebase — `CredibilityChart`, `MinerActivity`, `HomePage`, and the card-view donut all round. Flooring only the list-view column makes the same miner render 96% here and 97% everywhere else. Eligibility is computed server-side on the raw fraction (`isEligible`/`ossIsEligible`), so the displayed integer gates nothing and the rounding is cosmetic. Closing.

PR #1147, opened 2026-05-14
feat(dashboard): source Total Lines Committed

plind-junior:
## Summary

Replaces the client-side sum of `additions + deletions` over `mergedWindowPrs`
with the pre-aggregated `/dash/lines/hist-trend` time series from the dashboard
backend.

## Why

- **Single source of truth.** The dashboard backend already owns the canonical
  daily total. Computing the same value client-side risked drift if the
  backend's definition ever changed (e.g., excluding bot commits, filtering
  merge commits).
- **Less client work.** The KPI no longer requires materializing every PR's
  `additions + deletions` just to read one number.
- **Faster perceived load.** Hist-trend is a tiny payload (~one row per day)
  compared to walking the full PR set.

## Changes

- New `useLinesHistTrend()` hook in `DashboardApi.ts` consuming the existing
  `CommitsTrend` type (previously declared but unused).
- `buildDashboardKpis(prs, issues, linesHistTrend, range)` now accepts the
  trend array and sums only buckets inside the active window.
- `useDashboardData` calls the new hook and passes the data through.
- Other KPIs (Total Commits, Issues Solved, Total Repositories) still derive
  client-side — those have no equivalent backend endpoint.

## Test plan

- [ ] Compare displayed `Total Lines Committed` against the previous value for
  each preset window (1d / 7d / 35d / all) — should match.
- [ ] Verify the KPI updates when the time-range selector changes.
- [ ] Verify graceful fallback when the endpoint is empty / errors (KPI shows
  `0` instead of crashing).
- [ ] Confirm no regression in the other three KPI tiles.

image

### DIFF
--- a/src/api/DashboardApi.ts
+++ b/src/api/DashboardApi.ts
@@ -6,7 +6,6 @@ import {
   type Repository,
   type LanguageWeight,
   type CommitLog,
-  type CommitsTrend,
 } from './models/Dashboard';
 
 const useDashboardQuery = <TResponse = void, TSelect = TResponse>(
@@ -36,8 +35,11 @@ export const useReposAndWeights = () =>
 export const useLanguagesAndWeights = () =>
   useDashboardQuery<LanguageWeight[]>('useLanguagesAndWeights', '/languages');
 
-export const useLinesHistTrend = () =>
-  useDashboardQuery<CommitsTrend[]>('useLinesHistTrend', '/lines/hist-trend');
+export const useLinesTotal = (params: { from: string; to: string }) =>
+  useDashboardQuery<number>('useLinesTotal', '/lines/total', undefined, {
+    from: params.from,
+    to: params.to,
+  });
 
 export const useInfiniteCommitLog = (options?: {
   refetchInterval?: number;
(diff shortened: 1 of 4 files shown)


anderdc (changes requested):
The dashboard backend now has a `/dash/lines/total?from=&to=` endpoint (das-gittensor#74, deployed to prod) that returns the exact windowed `SUM(additions + deletions)` as a single number. Please switch this PR to use it instead of `hist-trend`.

`hist-trend` buckets by calendar day, so summing buckets client-side cannot match a sub-day rolling window — the `1d` preset and the `7d`/`35d` window edges drift from the old value, and it is inconsistent with the sibling "Total Commits" tile, which still windows precisely on PR timestamps.

Changes needed:
- `src/api/DashboardApi.ts`: replace `useLinesHistTrend()` with a hook that calls `/dash/lines/total`, passing `from`/`to` as ISO query params.
- `src/pages/dashboard/useDashboardData.ts`: derive `from`/`to` from `getWindowBounds(range)` and pass them to the hook.
- `src/pages/dashboard/dashboardData.ts`: `buildDashboardKpis` should take the scalar total directly — drop the `linesHistTrend` array param and the `filter`/`reduce`. The backend already windowed it.

Net result: the KPI matches exactly with no client-side windowing math.

anderdc (approved):

PR #1150, opened 2026-05-14
feat(search): match miner hotkey in global search

plind-junior:
## Summary

Lets users find a miner by their Bittensor hotkey in the global search bar (top of every page, dashboard included). Previously only matched against GitHub ID and GitHub username — pasting a hotkey returned no results.

Also adds a thin API hook for the unused `/miners/by-hotkey/{hotkey}/github` endpoint (from [api.gittensor.io swagger](https://api.gittensor.io/swagger)), available for future surfaces that need to resolve an off-list hotkey directly via the API.

### `src/api/models/Dashboard.ts`

Added the minimal binding type for the new endpoint:
​```ts
/** Minimal hotkey ↔ GitHub binding, returned by `/miners/by-hotkey/{hotkey}/github`. */
export type MinerHotkeyBinding = {
  uid: number;
  hotkey: string;
  githubId: string;
};
​```

### `src/api/MinerApi.ts`

New hook for direct hotkey resolution:
​```ts
export const useMinerGithubByHotkey = (hotkey: string, enabled?: boolean) =>
  useApiQuery<MinerHotkeyBinding>(
    'useMinerGithubByHotkey',
    `/miners/by-hotkey/${encodeURIComponent(hotkey)}/github`,
    undefined,
    undefined,
    enabled,
  );
​```

The hook is exported but isn't called anywhere yet — it's ready for future surfaces (wallet integration, chain explorer deep-links, "paste hotkey and navigate" flows). The in-memory matcher below covers every active miner without an extra request.

### `src/pages/search/searchData.ts`

Added `miner.hotkey` to the miner search candidates. The `MinerSearchData` shape already carried this field; `getMinerSearchResults` just wasn't consulting it.

​```diff
 const results = sortByMatchThenTiebreaker(
   miners,
   query,
-  (miner) => [miner.githubId, miner.githubUsername],
+  (miner) => [miner.githubId, miner.githubUsername, miner.hotkey],
   (miner) => miner.totalScore,
 );
​```

This goes through the same exact / startsWith / contains ranker as the existing name/ID matches, tiebroken by `totalScore` — no special hotkey-only code path.

### `src/components/layout/GlobalSearchBar.tsx`

Updated the placeholder so users know hotkey lookup is now supported:

​```diff
-placeholder="Search miners, repositories, PRs, issues..."
+placeholder="Search miners (name / hotkey), repositories, PRs, issues..."
​```

## Behavior

After this PR, typing or pasting any of the following into the global search bar surfaces the miner:

- Full hotkey: `5FXDTjpMqMrk2Gu1a2DPUHfjavwbWfhyX2QGXMbbFnuWmbVc` → exact match, top of Miners results
- Hotkey prefix: `5FXDTjpM` → starts-with match, all miners with that prefix
- Mixed-case partial: case-insensitive against the candidate list (same normalization as name/ID)
- Existing name or numeric GitHub ID: unchanged

The search is in-memory against the result of `useAllMiners()`. Every active miner with a registered hotkey is indexed and matchable without an extra API request.

## Test plan

- [ ] `npm run dev`, open the dashboard.
- [ ] Paste a real hotkey from `/leaderboard` into the global search → that miner appears as a Miners result.
- [ ] Type the first 8 characters of a hotkey → multiple miners with that prefix appear, sorted by `totalScore`.
- [ ] Existing name search (e.g. `opentensor`) and numeric GitHub-ID search still work.
- [ ] Placeholder text reads "Search miners (name / hotkey), repositories, PRs, issues..."

image

### DIFF
--- a/src/api/MinerApi.ts
+++ b/src/api/MinerApi.ts
@@ -8,6 +8,7 @@ import {
   type GithubMinerData,
   type MinerEvaluation,
   type CommitLog,
+  type MinerHotkeyBinding,
   type MinerIssue,
   type MinerIssuesResponse,
 } from './models/Dashboard';
@@ -72,6 +73,22 @@ export const useMinerGithubData = (githubId: string, enabled?: boolean) =>
   useMinersQuery<GithubMinerData>(
     'useMinerGithubData',
     `/${githubId}/github`,
+    undefined,
+    undefined,
+    enabled,
+  );
+
+/**
+ * Resolve a Bittensor hotkey to its GitHub binding. Returns `{ uid, hotkey,
+ * githubId }`. Use when the app has a hotkey from on-chain data or user input
+ * and needs to navigate to the corresponding miner profile.
+ *
+ * The hotkey field is case-sensitive (SS58). Pass it through as-is.
+ */
+export const useMinerGithubByHotkey = (hotkey: string, enabled?: boolean) =>
+  useApiQuery<MinerHotkeyBinding>(
+    'useMinerGithubByHotkey',
+    `/miners/by-hotkey/${encodeURIComponent(hotkey)}/github`,
     undefined,
     undefined,
     enabled,
--- a/src/api/MinerApi.ts
+++ b/src/api/MinerApi.ts
@@ -8,7 +8,6 @@ import {
   type GithubMinerData,
   type MinerEvaluation,
   type CommitLog,
-  type MinerHotkeyBinding,
   type MinerIssue,
   type MinerIssuesResponse,
 } from './models/Dashboard';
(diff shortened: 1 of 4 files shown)


anderdc (changes requested):
The hotkey search itself is the right call and it's correct — `searchData.ts` just needed `miner.hotkey` added to the candidate list, and `buildMinerSearchData` already populates that field. Scope it down to that:

Keep: `src/pages/search/searchData.ts` (the matcher change) and `src/components/layout/GlobalSearchBar.tsx` (the placeholder).

Drop: `useMinerGithubByHotkey` in `src/api/MinerApi.ts` and `MinerHotkeyBinding` in `src/api/models/Dashboard.ts` — both are unused, have no caller, and the in-memory matcher already covers every active miner without that endpoint.

anderdc (approved):

PR #1151, opened 2026-05-14
feat(miner-activity): add 7D/35D/All range selector to contribution heatmap

### LINKED ISSUE #1152: Miner Activity heatmap hardcodes "last 30 days"
The Developer Activity heatmap on the miner profile page currently only shows a single, fixed time window. The prominent number above the heatmap is always labeled "contribution(s) in the last 30 days" with no option to widen or narrow the view.

This creates two issues for users:

1. **No way to switch ranges.** Users who want to see a tighter recent view (e.g. the past week) or a fuller historical view (e.g. all-time activity) have no control to do so. Other parts of the app let users pick between short, medium, and full-history windows, so this page feels inconsistent.

2. **The "last 30 days" label can be misleading.** For miners who have less than 30 days of history, the prominent number still says "in the last 30 days" even though there isn't 30 days of activity to count - the caption directly underneath the heatmap reflects the real window (e.g. "in the last 17 day(s)"), so the two numbers visibly disagree.

**Repro:**

1. Open any miner profile → **Activity** tab.
2. Notice that there is no control to change the heatmap's time window.
3. Open a miner with fewer than 30 days of merged history and observe that the big number's "last 30 days" label does not match the smaller caption rendered directly below the heatmap.

Image

jack-stormentswe:
## Summary

The Developer Activity heatmap on every miner profile currently hardcodes a "last 30 days" window for its prominent count, while the dashboard's repository activity chart already supports a **7D / 35D / All** range selector ([`RepositoryPrActivityChart.tsx`](src/components/repositories/RepositoryPrActivityChart.tsx)). This PR brings the miner heatmap in line with that pattern — same option set, same `<Select>` styling, same `ActivityRange = '7d' | '35d' | 'all'` shape — so users can switch between recent and historical activity views without leaving the page.

## What changed

- Added a `7D / 35D / All` `<Select>` in the heatmap card header (matches dashboard styling exactly).
- Heatmap data + count are now driven by the selected range:
  - **7D** → 7-day window
  - **35D** → 35-day window (new default)
  - **All** → full window from earliest merged PR to today
- The prominent number's subtitle is now dynamic: `contribution(s) in the last 7 days` / `in the last 35 days` / `all time`.
- Window count is now consistent with the rendered cells (the previous "last 30 days" label counted 31 days due to an off-by-one — fixed as a side effect of replacing the magic constant).
- Renamed `ContributionHeatmap`'s `contributionsLast30Days` prop to the more honest `contributionsCount`, and added an optional `headerRight` slot so the dashboard's existing usage stays untouched.

## Why this approach

- Reuses an interaction pattern reviewers have already approved on the dashboard — no new design surface.
- Keeps the change scoped to two files (`MinerActivity.tsx` + `ContributionHeatmap.tsx`); no new shared utilities or routes.
- The default stays close to the old behavior (35D is the closest matching dashboard option to the old hardcoded 30) so existing users see a familiar view on first load.

## Verification

1. Open a miner profile → **Activity** tab.
2. Toggle the new selector through `7D`, `35D`, `All` and confirm:
   - The big number, subtitle, and heatmap cell count all stay consistent with the selected range.
   - The heatmap library's footer caption (`X contribution(s) in the last Y day(s)`) also matches.
3. Open a brand-new miner with very little history and select `All` — confirm the heatmap collapses to just the days they've been active.
4. Open the dashboard and confirm the repository activity chart's `<Select>` is visually identical to the new one on the miner page.

https://github.com/user-attachments/assets/88bec3e2-91bf-4d0a-a3fb-648ed04a75c5

Fixes #1152

### DIFF
--- a/src/components/ContributionHeatmap.tsx
+++ b/src/components/ContributionHeatmap.tsx
@@ -15,7 +15,7 @@ interface ContributionData {
 
 interface ContributionHeatmapProps {
   data: ContributionData[];
-  contributionsLast30Days: number;
+  contributionsCount: number;
   totalDaysShown: number;
   subtitle?: string;
   footerText?: string;
@@ -24,11 +24,12 @@ interface ContributionHeatmapProps {
   bare?: boolean;
   selectedDate?: string;
   onDayClick?: (date: string) => void;
+  headerRight?: React.ReactNode;
 }
 
 const ContributionHeatmap: React.FC<ContributionHeatmapProps> = ({
   data,
-  contributionsLast30Days,
+  contributionsCount,
   totalDaysShown,
   subtitle = 'network contribution(s) in the last 30 days',
   footerText,
@@ -37,6 +38,7 @@ const ContributionHeatmap: React.FC<ContributionHeatmapProps> = ({
   bare = false,
   selectedDate,
   onDayClick,
+  headerRight,
 }) => {
   const theme = useTheme();
   const heatmapLevels = [...CONTRIBUTION_HEATMAP_SCALE];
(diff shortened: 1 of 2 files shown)


davidv399:

anderdc:
Appreciate the work here, and the implementation itself is clean. After considering it, we'd rather keep the miner activity heatmap on a single fixed window — a range selector isn't a direction we want to take that view. This is a product decision, not a code issue. Closing.

PR #1157, opened 2026-05-14
fix(watchlist): prevent phantom miners from leaking into watchlist via /miners/details with an unknown id

### LINKED ISSUE #1159: [BUG] Watchlist can pin a miner that doesn't exist, leaving a phantom entry the user can't clean up
The miner detail page renders its "add to watchlist" star button as soon as a `githubId` is present in the URL, without first checking that the miner actually exists.

This means a user who lands on the miner detail page with any arbitrary value in the `githubId` query parameter - for example, by typing one in directly, by following a stale link to a deleted account, or by a redirect that loses its target - sees an active star button on top of the page's own "No data found" error block. Clicking the star pins the unknown id into the watchlist exactly as if it were a real miner.

Once an unknown id has been pinned, the watchlist behaves inconsistently:

- The watchlist sidebar badge and the per-category counts include the unknown entry in their totals.
- The watchlist's miners table only shows entries that resolve against the live list of tracked miners, so the unknown id is silently filtered out of the visible list.

The result is a watchlist where the badge says something like "Watchlist 2" while the visible list shows fewer items, and there is no in-app affordance to clear the missing entry. The user has to manually remove the value from `localStorage` to fix it.

The repository, PR, and bounty/issue detail pages already gate their watchlist buttons behind a successful data fetch - only the miner detail page is missing this guard.

**Expected behavior:**

- The star button on the miner detail page should not be available until the underlying miner has been confirmed to exist. While the data is still loading, or when the fetch resolves with no record, the star should be hidden (or disabled) so the user can't pin something the rest of the app considers nonexistent.
- The watchlist itself should remain internally consistent: any watched entry that can't be resolved against the live data should be visible to the user with a way to remove it, so the counts in the sidebar always match what the user can actually see and act on.

**Repro:**

1. Open `/miners/details?githubId=123456789` (or any other value not present on the leaderboard).
2. Observe that the page renders the "No data found for GitHub user…" error block, but the star button at the top of the page is still active.
3. Click the star, then navigate to `/watchlist`.
4. Observe that the sidebar shows a watchlist count that includes the unknown entry, while the miners tab shows no row for it and offers no way to clean it up.

https://github.com/user-attachments/assets/be2ea6ec-847f-42af-939b-7fd994aacdce

jack-stormentswe:
Closes #1159

## Summary

The Miner Details page renders the "add to watchlist" star button as soon as a `githubId` is present in the URL - before the underlying miner data has been fetched. This makes it possible to pin a miner that doesn't actually exist (e.g. by manually visiting `/miners/details?githubId=123456789`), which leaks a phantom entry into the watchlist that can never be opened to anything but a "No data found" error page.

It also produces a visible inconsistency in the watchlist itself: the sidebar/category badge counts every id in storage, but the miners table inner-joins against `useAllMiners()` and silently drops anything that doesn't resolve - so the badge can read "Watchlist 2" while the visible list shows zero or one miner, with no way for the user to clean up the difference.

The repository and PR/issue detail pages already gate their watchlist buttons behind a successful data fetch - only the miner details page was missing this guard.

## Changes

- [`src/pages/MinerDetailsPage.tsx`](src/pages/MinerDetailsPage.tsx) - call `useMinerStats(githubId)` at the page level and render `<WatchlistButton>` only after the fetch resolves with a real record. While loading, or when the miner is unknown, the star is hidden. (TanStack Query dedupes against the existing call inside `MinerScoreCard`, so this adds no extra network traffic.)
- [`src/pages/WatchlistPage.tsx`](src/pages/WatchlistPage.tsx) - `MinersList` now detects unresolved watched ids (entries that aren't in the all-miners cache once it's loaded) and surfaces them as a warning card with a one-click **Remove unresolved** action. The visible list is unchanged for resolved miners.

## Verification

1. Open `/miners/details?githubId=123456789` in an incognito window. While the request is in flight the star is absent; once the API responds with no record the star stays hidden and the page collapses to the existing "No data found" card. The star can no longer be clicked.
2. Open a real miner profile (e.g. via the leaderboard). The star renders as before after the page data loads, and toggling it adds/removes the miner from the watchlist exactly as it did previously.
3. Pre-seed a phantom watchlist entry from DevTools - `localStorage.setItem('gittensor.watchlist.v2', JSON.stringify({ miners: ['999999'], repos: [], bounties: [], prs: [], issues: [] }))` - then reload `/watchlist`. The sidebar badge reads "1", and the miners tab now shows a warning card explaining the entry couldn't be loaded with a **Remove unresolved** button that clears it. Clicking it removes the entry from localStorage and the badge updates to "0".
4. The unresolved card only appears once `useAllMiners()` has finished loading, so a slow network won't cause a false positive on first paint.

## Recording

https://github.com/user-attachments/assets/819e8c7b-c48e-4654-8898-3122b8107195

### DIFF
--- a/src/pages/MinerDetailsPage.tsx
+++ b/src/pages/MinerDetailsPage.tsx
@@ -15,6 +15,7 @@ import {
   SEO,
 } from '../components';
 import { WatchlistButton } from '../components/common';
+import { useMinerStats } from '../api';
 
 type ViewMode = 'prs' | 'issues';
 
@@ -84,6 +85,13 @@ const MinerDetailsPage: React.FC = () => {
     setSearchParams(p, { replace: true });
   };
 
+  const { data: minerStats, isLoading: isLoadingMinerStats } = useMinerStats(
+    githubId ?? '',
+  );
+  // Only show the star once we've confirmed the miner exists — otherwise
+  // users can pin phantom entries via /miners/details?githubId=<anything>.
+  const minerExists = !isLoadingMinerStats && !!minerStats;
+
   if (!githubId) {
     return <Navigate to="/top-miners" replace />;
   }
@@ -131,11 +139,13 @@ const MinerDetailsPage: React.FC = () => {
               }}
             >
               <BackButton to="/top-miners" mb={0} />
-              <WatchlistButton
-                category="miners"
-                itemKey={githubId}
-                size="medium"
-              />
+              {minerExists && (
+                <WatchlistButton
+                  category="miners"
+                  itemKey={githubId}
+                  size="medium"
+                />
+              )}
             </Box>
             <Box
               sx={{
--- a/src/pages/WatchlistPage.tsx
+++ b/src/pages/WatchlistPage.tsx
@@ -919,6 +919,7 @@ const WatchlistOptionsButton: React.FC<WatchlistOptionsButtonProps> = ({
 
 const MinersList: React.FC<{ itemKeys: string[] }> = ({ itemKeys }) => {
   const { data: allMinersStats, isLoading } = useAllMiners();
+  const { remove } = useWatchlist('miners');
   const watchedSet = useMemo(() => new Set(itemKeys), [itemKeys]);
 
   const minerStats = useMemo(() => {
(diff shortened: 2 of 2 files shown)


davidv399:

anderdc (approved):

PR #1160, opened 2026-05-14
fix(back-button): require same-origin referrer before stepping browser history, so fresh-tab Back falls back to the in-app route

### LINKED ISSUE #1162: Shared "Back" button on detail pages can navigate outside the app when the page was opened from a fresh tab or shared link
The shared back-navigation button used on every detail page in the app (Miner, Repository, PR, Issue, and the Search results page) decides whether to step the browser history or fall back to a known in-app route based on a single signal - whether the browser reports more than one entry in its history stack.

That signal is not a reliable indicator that the previous history entry actually belongs to the app. In several common flows the browser reports a longer history even though nothing before the current page came from inside the app:

- The user opens a shared deep link in a new tab - the new-tab page itself counts as a history entry.
- The user opens the page from a bookmark or by pasting a URL into the address bar.
- The user follows an external link (from social media, chat, email, etc.) that lands directly on a detail page.

In all of these flows, clicking the "Back" button on the destination detail page takes the user *out of the app* - back to the new-tab page, the previous external site, or wherever the browser's previous entry pointed - instead of routing them to the sensible in-app list page (top miners, repositories, bounties, etc.) the button is supposed to be able to fall back to.

The same root-cause issue was recently addressed on the 404 page's own "Go back" button, but the broader fix needs to live in the shared back-button component so every detail page that uses it picks it up.

**Expected behavior:**

- "Back" should only step the browser history when the previous history entry is verifiably an in-app page.
- When the previous entry can't be confirmed to be in-app, the button should fall back to a known route - preferably the per-page parent route the component already accepts as a fallback (so a deep-linked miner page goes to the miners list, a deep-linked repo page goes to the repositories list, etc.), rather than dumping the user on the home page.
- A one-click "Back" should never silently take the user out of the app.

**Repro:**

1. Open a detail page directly in a fresh browser tab (e.g. paste `/miners/details?githubId=…` into the address bar of a brand-new tab, or click a link to it from a different site / app).
2. Without navigating anywhere else first, click the "Back" button at the top of the page.
3. Observe that the browser exits the app - landing on the new-tab page, the previous external site, or whatever the browser's prior history entry pointed at - instead of the sensible in-app list page.
4. Reproduce on the other detail surfaces (`/miners/repository`, `/miners/pr`, `/bounties/details`, `/search`) by opening any of them in a new tab the same way.

**How to Verify:**

https://github.com/user-attachments/assets/f45484aa-50a4-4b20-9132-2a42028ada7a

pandadev66:
## Summary

The shared [`BackButton`](src/components/BackButton.tsx) component used on every detail page (Miner, Repository, PR, Issue, Search) decides whether to step the browser history or fall back to a known in-app route based solely on `window.history.length > 1`. That check isn't a reliable signal that the previous history entry actually belongs to the app — opening a detail page directly in a fresh tab, from a bookmark, or by clicking a shared link from outside the app all leave a non-app entry as the "previous" page, so clicking "Back" calls `navigate(-1)` and the user is taken out of the app.

This PR tightens the check with the same approach used for the 404 page: require `document.referrer` to share the app's origin before stepping back. When the previous entry can't be confirmed in-app, the button falls through to the per-page `to` route (e.g. `/top-miners` from a miner profile, `/repositories` from a repo page) that the component already accepts as a fallback.

## Behavior

- **Navigated in from another app page** (`history.length > 1` AND same-origin `referrer`) → `navigate(-1)` — unchanged from today.
- **Router state carries an explicit `backTo`** (set by list pages when they link into a detail page, e.g. "Back to Watchlist") → `navigate(state.backTo)` — unchanged.
- **Fresh tab / bookmark / external link** (no referrer or cross-origin referrer) → `navigate(to)` — the page-specific in-app fallback the component already takes as a prop. Previously this fell through to `navigate(-1)` and left the app.

## Verification

1. Open `/miners/details?githubId=<known-id>` directly in a brand-new tab (paste the URL into the address bar, or click a link to it from a non-app origin). Click "Back" once — the app should land on `/top-miners` rather than the browser's previous tab/page.
2. Repeat for the other detail pages by opening them the same way: `/miners/repository?name=…` → `/repositories`, `/miners/pr?repo=…&number=…` → `/repositories`, `/bounties/details?id=…` → `/bounties`.
3. Navigate normally from a list page to a detail page (e.g. click a row on `/top-miners`). The "Back" button should still call `navigate(-1)` and return to the exact list state, including any filters or scroll the page restores from router state.
4. Set an explicit `backTo` in router state (e.g. "Back to Watchlist") and confirm the button still routes to that destination when the in-app history signal isn't trusted.

## Why this approach

- One small change to a shared component fixes the symptom on every page that consumes it; no per-page changes needed.
- Mirrors the same-origin referrer check that's already been deployed on the 404 page's own "Go back" button, so the two surfaces behave consistently.
- The `to` prop already exists on every caller (it's required), so there's no new copy or routing config to keep in sync — the fix uses what's already there.

## Recording

https://github.com/user-attachments/assets/0a474b6a-a14e-45eb-b8f0-454b3cf50eb2

Fixes #1162

### DIFF
--- a/src/components/BackButton.tsx
+++ b/src/components/BackButton.tsx
@@ -21,7 +21,21 @@ const BackButton: React.FC<BackButtonProps> = ({
   const location = useLocation();
   const state =
     (location.state as { backLabel?: string; backTo?: string }) || {};
-  const canGoBack = typeof window !== 'undefined' && window.history.length > 1;
+  // `history.length > 1` alone isn't a reliable signal that the previous
+  // entry is an in-app page — opening a deep link in a fresh tab counts
+  // the browser's own previous entry, so `navigate(-1)` would leave the
+  // app. Also require the referrer to be same-origin before stepping back.
+  const canGoBack = (() => {
+    if (typeof window === 'undefined') return false;
+    if (window.history.length <= 1) return false;
+    const ref = document.referrer;
+    if (!ref) return false;
+    try {
+      return new URL(ref).origin === window.location.origin;
+    } catch {
+      return false;
+    }
+  })();
   const displayLabel = state.backLabel ?? label;
 
   const handleClick = () => {


pandadev66:

anderdc:
fix conflicts

pandadev66:
> fix conflicts

Done ❤️

pandadev66:
Hi @anderdc 
When you have a moment, please take a look at this PR. ❤️

anderdc:
`document.referrer` is set once at document load and never changes across in-app React Router navigation, so it cannot distinguish an in-app previous page from an external one. Anyone who opens the site fresh (empty referrer) and then browses in-app gets `canGoBack === false` on every detail page, so Back calls `navigate(to)` instead of `navigate(-1)` — on a repo → PR path that lands on the generic repositories list, not the actual previous page. That regresses normal Back navigation. The fresh-tab case this targets is also already handled: a freshly opened tab has `history.length === 1`, so the existing check already falls back to `navigate(to)`. Closing.

PR #1174, opened 2026-05-14
fix: show in-view rank indicator when leaderboard filters are active (#1102)

### LINKED ISSUE #1102: [UX] Leaderboard rank column shows global ranks with unexplained gaps under any filter
## Description
On `/top-miners` (and `/discoveries`), the **Rank** column always shows each miner's **global** rank across the entire dataset. When a filter is applied - most visibly `Eligibility = Eligible` - the visible rows are a subset of the global list, but the rank column keeps the global numbers. The visible result is a long sequence of unexplained jumps in the rank column (e.g. `16, 19, 23, 48, 94`) with no in-UI indication that the gaps are caused by filtered-out miners.

The math is correct (the missing ranks correspond to ineligible miners), but to a user the column reads as "buggy or skipping data." This is a discoverability / clarity defect rather than a correctness one.

## Steps to Reproduce
1. Open `/top-miners` (or `/discoveries`).
2. Switch to **list view** (the table icon in the Filters → View section) so the Rank column is visible.
3. Expand **Filters** in the right sidebar.
4. Set **Eligibility** to `Eligible`.
5. Scroll the list of eligible miners.
6. Observe the Rank column: consecutive entries near the top (`7, 8, 9, 10, 11, 12, 13, 14, 15, 16`) followed by sudden jumps (`19, 23, 48, 94`) further down. The same pattern appears in the OSS Contributions and Discoveries timelines.

## Expected Behavior
One of the following should be true, and the choice should be discoverable from the UI:

**Option A (recommended) - show two ranks**: keep the global rank in its current column, and add a small "Rank in this view" indicator (e.g. an unobtrusive `· 11 in view` next to the global rank, or a secondary column when any filter is active). This preserves the global context while making the filtered list feel consistent.

**Option B - make global rank discoverable**: keep the current behavior but add a tooltip on the Rank column header and/or a one-line note above the table when filters are active: "Rank shows global position across all 108 miners - gaps indicate miners hidden by the current filter."

**Option C - switch to local rank under filters**: when any filter is active, renumber rows `1, 2, 3, …` within the filtered subset. Simplest but loses the "what's their actual standing" signal.

## Actual Behavior
- The Rank column always shows the global rank.
- There is no tooltip, header annotation, or status line explaining the gaps.
- Users reasonably interpret `16 → 19 → 23 → 48 → 94` as a bug.

## Notes
This is purely a presentation question. No backend changes, no sort/filter logic changes - only how the rank column communicates context.

## File Reference
- `src/components/leaderboard/TopMinersTable.tsx` - list view row rendering (Rank column).
- Same component handles both `OSS Contributions` and `Discoveries` timelines, so the fix applies to both.

Image

alpurkan17:
## Summary

The Rank column on Top Miners / Discoveries leaderboards always shows the global rank. When a filter is active, consecutive rows show gaps (16, 19, 23, 48, 94) with no UI indication that the gaps are caused by filtered-out miners. Users reasonably interpret this as a bug.

Fix (Option A): add a subtle "·N in view" indicator next to the global rank when filters are active and the two ranks diverge.

## Changes

- **types.ts**: add `viewRank` to `MinerStats` — 1-indexed position in filtered subset
- **TopMinersTable.tsx**: compute `viewRank` only when a search/eligibility filter is active
- **MinersList.tsx**: show `·{viewRank}` beside the RankIcon in list view
- **MinerCard.tsx**: show `·{viewRank} in view` next to the global rank on cards

## Edge cases

- No filter active → no indicator shown
- First row (rank=viewRank=1) → no indicator shown
- Pure search filter → indicator shown for mismatched rows
- Both search + eligibility → indicator shown correctly
- No backend changes — computed entirely in frontend

## Validation

- [x] 4 files, +52/−6, no new dependencies
- [x] Null-guarded: `viewRank != null` before comparison
- [x] Both card view and list view covered

Closes #1102

### DIFF
--- a/src/components/leaderboard/MinerCard.tsx
+++ b/src/components/leaderboard/MinerCard.tsx
@@ -219,11 +219,26 @@ export const MinerCard: React.FC<MinerCardProps> = ({
                   opacity: isEligible ? 1 : INACTIVE_OPACITY,
                   lineHeight: 1,
                   flexShrink: 0,
-                })}
-              >
-                #{miner.rank}
-              </Typography>
-            </Box>
+                 })}
+               >
+                 #{miner.rank}
+               </Typography>
+               {miner.viewRank != null && miner.viewRank !== miner.rank && (
+                 <Typography
+                   component="span"
+                   sx={(theme) => ({
+                     fontFamily: FONTS.mono,
+                     fontSize: '0.58rem',
+                     fontWeight: 500,
+                     color: theme.palette.text.tertiary,
+                     lineHeight: 1,
+                     flexShrink: 0,
+                   })}
+                 >
+                   ·{miner.viewRank} in view
+                 </Typography>
+               )}
+             </Box>
             {/* Badges under the name */}
             <Box
               sx={{
--- a/src/components/leaderboard/MinersList.tsx
+++ b/src/components/leaderboard/MinersList.tsx
@@ -120,7 +120,24 @@ export const MinersList: React.FC<MinersListProps> = ({
       header: 'Rank',
       width: '60px',
       cellSx: { pr: 0 },
-      renderCell: (miner) => <RankIcon rank={miner.rank ?? 0} />,
+      renderCell: (miner) => (
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 0.75 }}>
+          <RankIcon rank={miner.rank ?? 0} />
+          {miner.viewRank != null && miner.viewRank !== miner.rank && (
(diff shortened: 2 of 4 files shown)


alpurkan17:
@anderdc @landyndev ready for review

anderdc:
PRs must target the **test** branch, not main — see [CONTRIBUTING.md](https://github.com/entrius/gittensor-ui/blob/test/CONTRIBUTING.md). Closing.

PR #1196, opened 2026-05-15
fix: prevent BackButton from navigating outside the app

### LINKED ISSUE #1162: Shared "Back" button on detail pages can navigate outside the app when the page was opened from a fresh tab or shared link

zp6:
Fixes #1162

The shared BackButton used on every detail page now checks document.referrer to determine whether the previous history entry belongs to this app. When the referrer is missing or points to a different origin (fresh tab, shared link, bookmark), the button falls back to the configured in-app route instead of calling 
avigate(-1) which would leave the app.

**Changes:**
- Added isInAppNavigation() helper that checks document.referrer origin against window.location.origin`n- BackButton now only steps through browser history when the previous entry is verifiably an in-app page
- Falls back to state.backTo or the 	o prop when the referrer is empty/external

**Testing:**
- Open a detail page from a fresh tab or shared link ? Back button navigates to the parent list page
- Navigate within the app to a detail page ? Back button correctly goes to the previous in-app page
- Verified no regression in normal in-app navigation flows

### DIFF
--- a/src/components/BackButton.tsx
+++ b/src/components/BackButton.tsx
@@ -12,6 +12,27 @@ interface BackButtonProps {
   mb?: number;
 }
 
+/**
+ * Determine whether the previous history entry likely belongs to this app.
+ * We consider it safe to go back only when we have an explicit navigation
+ * state marker (`fromApp`) set by internal links, **or** when the
+ * `document.referrer` points to the same origin.
+ */
+function isInAppNavigation(): boolean {
+  // If the previous navigation was performed by react-router (pushState),
+  // the referrer stays the current origin. For truly external arrivals
+  // (shared link, fresh tab, bookmark) the referrer is either empty or a
+  // different origin.
+  try {
+    const referrer = document.referrer;
+    if (!referrer) return false; // direct navigation, no referrer
+    const refUrl = new URL(referrer);
+    return refUrl.origin === window.location.origin;
+  } catch {
+    return false;
+  }
+}
+
 const BackButton: React.FC<BackButtonProps> = ({
   label = 'Back',
   to,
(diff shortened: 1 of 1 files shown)


anderdc:
Duplicate of #1160 which is better scoped/implemented. Closing.

PR #1197, opened 2026-05-15
fix: auto-recover from stale lazy-chunk hash after redeploys

zp6:
Fixes #1167

When a user has a long-lived tab from before a production deploy, navigating to a lazy route fails with 'Failed to fetch dynamically imported module' because the old entry chunk references deleted hashed filenames.

**Changes:**
- Introduced lazyWithReload() wrapper that replaces every React.lazy() call in src/routes.tsx`n- On dynamic import failure, triggers a single window.location.reload() to re-fetch the current entry HTML
- Returns a never-resolving promise so React Suspense stays mounted during the reload

**Root cause:** Vite emits hashed chunk filenames (e.g. WatchlistPage-C10WDGR6.js). After a new deploy, old chunks are deleted. Users with stale tabs hit 404s on dynamic imports.

**Verification:**
- Simulated stale chunk by renaming a built asset ? navigating triggers reload ? page loads successfully
- 
pm run build passes

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -7,31 +7,55 @@ export type AppRoute = Omit<PathRouteProps, 'path'> & {
   showGlobalSearch?: boolean;
 };
 
+/**
+ * Wrap React.lazy so that a failed dynamic import (e.g. stale chunk hash
+ * after a production deploy) triggers a single page reload instead of an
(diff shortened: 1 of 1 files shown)


anderdc:
Duplicate of #1167 which is better scoped/implemented. Closing.

PR #1108, opened 2026-05-13
fix(miners): debounce search in MinerPRsTable and MinerOpenDiscoveryIssuesByRepo

cleanjunc:
## Summary

Two miner views were missed by the `DebouncedSearchInput` migration in #1084 and still committed `searchQuery` on every keystroke, forcing the parent and its filter/sort/paginate `useMemo` chains to rerun per character.

* `MinerPRsTable.tsx`: wrap the PR search field in `DebouncedSearchInput`. The `setPage(0)` reset that used to ride along with the keystroke handler now lives in the debounced commit path, gated by a `searchQueryRef` comparison so the wrapper's mount fire and the existing `[githubId]` reset do not clobber a deep-linked `?prPage=N` URL.
* `MinerOpenDiscoveryIssuesByRepo.tsx`: wrap the search field inside `renderToolbar`, so the "Your open discovery issues" and "Other open discovery issues" sections each get an independent debounced draft. Typing in one section no longer rerenders the other.

Both files now match the existing pattern used by `MinerRepositoriesTable`, `IssuesList`, `TopRepositoriesTable`, `MinerScoreBreakdown`, `LanguageWeightsTable`, and `WatchlistPage`.

## Related Issues

Fixes #1107

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Refactor
* [ ] Documentation
* [ ] Other (describe below)

## Screenshots

Before:
[before.webm](https://github.com/user-attachments/assets/d5bad137-c2b7-491b-93c8-547793bb7541)

After:
[after.webm](https://github.com/user-attachments/assets/d69e5fbd-e838-4a0c-90d7-cae41fc011fe)

## Checklist

* [x] New components are modularized/separated where sensible
* [x] Uses predefined theme (e.g. no hardcoded colors)
* [x] Responsive/mobile checked
* [x] Tested against the test API
* [x] `npm run format` and `npm run lint:fix` have been run
* [x] `npm run build` passes
* [x] Screenshots included for any UI/visual changes

PR #1246, opened 2026-05-19
fix(repositories): keep deep-linked `?page=N` when search hasn't changed

### LINKED ISSUE #1247: [Bug] /repositories pagination snaps back to page 1 after every click and on direct URL load
## Description

On `/repositories`, clicking "next page" or "last page" in the repositories table advances for one render and then snaps back to page 1. Direct loads of `/repositories?page=N` strip `?page` on first paint and render page 1 instead. The bug only affects `/repositories`; other tables using the same search input do not snap back.

## Steps to Reproduce

1. Open `/repositories` (default list view, 10 rows). The current registry has >10 entries so page 2 exists.
2. Click the "next page" or "last page" button.
3. Observe the URL briefly become `?page=1`, then revert to bare. The table re-renders page 1.

Also reproducible by loading `/repositories?page=1` directly.

## Expected Behavior

Page advances and the URL retains `?page=N` until another navigation action.

## Actual Behavior

`?page=` is removed from the URL within one render tick and the table returns to page 1.

## Root Cause

`src/components/common/DebouncedSearchInput.tsx:67-69` runs the effect on `onDebouncedChange` identity changes, not only on `debounced` value changes. In `TopRepositoriesTable`, `onDebouncedChange` routes through `useDataTableParams.setFilter`, whose `useCallback` depends on `react-router-dom`'s `setSearchParams` — a ref that changes on every URL update. So every `setPage` triggers `setFilter('search', '')`, which deletes `?page` (its `resetPageOnChange` default is `true`).

## Suggested Fix

Stabilize the callback in a ref and skip the initial-mount fire so direct `?page=N` loads aren't stripped:

```ts
const callbackRef = useRef(onDebouncedChange);
useEffect(() => {
  callbackRef.current = onDebouncedChange;
});

const didFireInitialRef = useRef(false);
useEffect(() => {
  if (!didFireInitialRef.current) {
    didFireInitialRef.current = true;
    return;
  }
  callbackRef.current(debounced);
}, [debounced]);

RenzoMXD:
## Summary

The repositories leaderboard's pagination snaps back to page 1 after every click and on direct loads of `/repositories?page=N`. The table routes `DebouncedSearchInput`'s `onDebouncedChange` through `useDataTableParams.setFilter('search', ...)`, whose `useCallback` closes over `react-router-dom`'s `setSearchParams`. That ref is recreated on every URL change, so every `setPage(N)` recreates the callback, which re-fires the wrapper's mount-anchored debounce effect. The fired commit deletes `?page` via the filter's default `resetPageOnChange`, snapping the table back to page 1 right after the click.

Same root cause #1108 fixed for `MinerPRsTable`. This PR applies the same callsite gate: keep the latest committed `searchQuery` in a ref and skip `setFilter('search', ...)` when the wrapper hands back the value we already hold. Typing a new search still commits normally and resets `?page` as intended; the mount fire and any re-render driven by an unrelated URL change no longer touches the page slot.

Closes #1247

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

Verified locally against `npm run dev`:

- Direct load `/repositories?page=1` retains `?page=1` and renders rows 13–15 of 15.
- "Go to next page" / "Go to last page" advance to `?page=1` and stay (no snap-back).
- Typing into the search input still commits the filter after the debounce and resets `?page` so the result list starts at page 1.
- `npm run build` — clean
- `npm run lint` — `--max-warnings 0` clean
- `npx prettier --check` — clean

## Checklist

- [x] Responsive/mobile checked (behavior is URL/query only)
- [x] `npm run build` passes

## Screenshots

Before:

https://github.com/user-attachments/assets/954b73cf-c2d0-4d34-9963-4864a8f28216

After:

https://github.com/user-attachments/assets/a05846fa-dd5a-42bb-a8d2-5d09160ec3bd

### DIFF
--- a/src/components/leaderboard/TopRepositoriesTable.tsx
+++ b/src/components/leaderboard/TopRepositoriesTable.tsx
@@ -1,4 +1,10 @@
-import React, { useState, useMemo, useCallback } from 'react';
+import React, {
+  useState,
+  useMemo,
+  useCallback,
+  useRef,
+  useEffect,
+} from 'react';
 import {
   Box,
   Card,
@@ -256,8 +262,17 @@ const TopRepositoriesTable: React.FC<TopRepositoriesTableProps> = ({
   const [showChart, setShowChart] = useState(false);
   const [useLogScale, setUseLogScale] = useState(true);
 
+  const searchQueryRef = useRef(searchQuery);
+  useEffect(() => {
+    searchQueryRef.current = searchQuery;
+  });
+
   const handleSearchChange = useCallback(
-    (value: string) => setFilter('search', value),
+    (value: string) => {
+      if (value === searchQueryRef.current) return;
+      searchQueryRef.current = value;
+      setFilter('search', value);
+    },
     [setFilter],
   );
 


anderdc:
#1249 already fixed this at the source — `setFilter` in `useDataTableParams` now clears the page slot only when the search value actually changed, so a re-emitted unchanged value no longer drops `?page` on a click or a direct `?page=N` load. This callsite ref-gate duplicates that check. Closing.

PR #1266, opened 2026-05-23
feat(bounties): add inline search to mobile toolbar in IssuesList

### LINKED ISSUE #808: Responsive: /bounties table toolbar is crowded on mobile (375px)
## Description

On `/bounties`, the table toolbar becomes crowded on mobile width (`375px`), making filter chips and controls hard to read/use.

## Steps to Reproduce

1. Open `https://gittensor.io/bounties`
2. Enable mobile responsive mode at width `375px`
3. Scroll to the bounties table controls area (filters + chart toggle + rows + search)

## Expected Behavior

- Toolbar should stack into clear rows on narrow screens
- Filter chips should wrap with consistent spacing
- Search input and rows selector should keep comfortable width and touch targets
- No crowding/overlap at common mobile widths (375px)

## Actual Behavior

- Filter chips and controls are tightly packed
- Rows selector and search are cramped
- Overall toolbar appears visually compressed in one block

## Environment

- Browser: Chrome
- OS: Windows 11

## Additional Context

Image

nickmopen:
## Summary

  On screens below the `xl` breakpoint (< 1536px), the `/bounties` toolbar only showed an Options button — search was hidden inside a popover requiring an extra tap. This PR adds an
  always-visible inline search field to the toolbar so users can filter issues without opening the popover.

  **Changes in `src/components/issues/IssuesList.tsx`:**
  - Moved `usePortal` declaration above `optionsPortalFilters` so the search block can be conditionally rendered
  - `optionsPortalFilters`: search section now renders only when `usePortal` is true (xl+ sidebar), removing the duplicate entry from the mobile popover
  - `inlineToolbar`: added `IssuesListSearchTextField` filling available width, with the Options button pinned right
  - `return`: moved `<Card>` inside `<DebouncedSearchInput>` so the inline search field shares the same search context

  ## Related Issues

  Closes #808

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor
  - [ ] Documentation
  - [ ] Other (describe below)

  ## Screenshots

  

  **Before** — only the Options button visible in the toolbar:

Screenshot 2026-05-23 074037

  **After** — inline search field always visible next to Options button:

 
Screenshot 2026-05-23 074052

  ## Checklist

  - [x] `npm run build` passes

### DIFF
--- a/src/components/issues/IssuesList.tsx
+++ b/src/components/issues/IssuesList.tsx
@@ -1015,6 +1015,8 @@ const IssuesList: React.FC<IssuesListProps> = ({
     </Box>
   );
 
+  const usePortal = !!portalTarget && isLargeScreen;
+
   const optionsPortalFilters = (
     <>
       {sortSection}
@@ -1023,10 +1025,12 @@ const IssuesList: React.FC<IssuesListProps> = ({
         {viewToggle}
       </Box>
       {rowsSection}
-      <Box>
-        <Typography sx={sidebarLabelSx}>Search</Typography>
-        <IssuesListSearchTextField fullWidth />
-      </Box>
+      {usePortal && (
+        <Box>
+          <Typography sx={sidebarLabelSx}>Search</Typography>
+          <IssuesListSearchTextField fullWidth />
+        </Box>
+      )}
       <Box>
         <Typography sx={sidebarLabelSx}>Chart</Typography>
         <Stack direction="row" alignItems="center" spacing={1}>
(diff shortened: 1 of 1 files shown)


ventura-oss:
Thanks for the PR. After looking at the before/after, I don't think adding an inline search field is the right fix here.

The page already has a search bar at the top of the layout. Adding a second visually-identical search input inside the toolbar means the same view ends up with two search bars stacked one above the other, which is more confusing than the original.

The real issue with the sub-`xl` toolbar isn't that it lacks a search — it's that the Options button is right-aligned all by itself, which leaves a whole row of dead space to its left. This PR is filling that dead space rather than addressing it.

A better direction would be to reposition Options so the empty row doesn't exist in the first place — for example, inline with the tab row (All / Available / Pending / History), or grouped with the other header/toolbar controls — rather than introducing a duplicate search affordance to justify the row.

Happy to look at a follow-up that takes that approach.

PR #1272, opened 2026-05-25
refactor(common): extract shared LoadingCard component

fansilas:
## Summary

Adds a shared `<LoadingCard>` component at `src/components/common/LoadingCard.tsx` to replace the inline `<Card sx={{ p: 4, textAlign: 'center' }} elevation={0}>` + `<CircularProgress>` pattern that was duplicated across several loading states.

The component accepts an optional `size` (defaults to 40) and an optional `sx` that merges over the base Card styles, so callers with custom backgrounds or borders can still use it.

```tsx
interface LoadingCardProps {
  size?: number;
  sx?: SxProps<Theme>;
}
```

Replaced inline usages:

* `src/components/miners/MinerIdentityRail.tsx` (size 32)
* `src/components/miners/MinerOpenDiscoveryIssuesByRepo.tsx`
* `src/components/miners/MinerPRsTable.tsx`
* `src/components/prs/PRDetailsCard.tsx` (custom sx forwarded for background and border)

Sites that wrap the spinner in a `<Box>` rather than a `<Card>` are a different layout (centered spinner in a flex container, no card chrome) and were intentionally left alone. The two `RepositoryPRsTable` / `RepositoryIssuesTable` loading states render a section header alongside the spinner, so they do not match the spinner-only `<LoadingCard>` contract and were also left as is to preserve current behavior.

## Related Issues

N/A

## Type of Change

* [ ] Bug fix
* [ ] New feature
* [x] Refactor
* [ ] Documentation
* [ ] Other (describe below)

## Screenshots

No visual change is expected: the spinner renders at the same size, position, and color on every replaced site.

## Checklist

* [x] New components are modularized/separated where sensible
* [x] `npm run build` passes
* [ ] Screenshots included for any UI/visual changes


anderdc (changes requested):
Clean extraction — the four replacements are behavior-equivalent and the array-sx merge in LoadingCard is correct, including the forwarded background/border in PRDetailsCard.

One thing to fix: MinerOpenDiscoveryIssuesByRepo.tsx imports LoadingCard via the direct path `from '../common/LoadingCard'`, while the other three sites import it through the barrel (`from '../common'`). Since you added the export to common/index.ts, route this site through the barrel too for consistency: change `import { LoadingCard } from '../common/LoadingCard';` to `from '../common'`.

fansilas:
Thank you for the review, @anderdc 
I updated it. Can you check again please?

anderdc (approved):

PR #1275, opened 2026-05-26
fix(watchlist): repos chart bars render at zero due to wrong config field

blinkeye-lcm:
## Summary

The bar chart on the Watchlist repositories tab renders every bar with value `0`, contradicting the `Weight` column shown in the list view directly beneath it. The chart reads `repo.config?.weight`, but the API config blob has no `weight` field, only `emissionShare`. The access typechecks because `RepositoryConfig` carries a `[key: string]: unknown` index signature, so the value is always `undefined` at runtime and every bar collapses to zero.

Fixed by reading from the already hoisted `repo.weight` field on `WatchedRepoStats`, which was added specifically for downstream sort and render code per the type's own comment at [src/pages/WatchlistPage.tsx:1020-1030](src/pages/WatchlistPage.tsx#L1020-L1030).

### Reproducer

1. Run `npm run dev`
2. Star at least two repositories from the watchlist controls
3. Navigate to `/watchlist?tab=repos`
4. Open the Options panel and enable the chart view via the BarChart icon
5. Every bar renders at zero. With the default log scale they sit below the axis floor and are invisible. With linear scale they render as flat zero-height bars

## Related Issues

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

**Before:** chart toggled on with two pinned repos, every bar at zero (invisible under log scale).

b1

**After:** chart toggled on with the same two pinned repos, bars render with values matching the `Weight` column in the list view below.

a1

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/pages/WatchlistPage.tsx
+++ b/src/pages/WatchlistPage.tsx
@@ -1655,7 +1655,7 @@ const ReposList: React.FC<{ itemKeys: string[] }> = ({ itemKeys }) => {
     const chartFont = echartsFontFamily(theme);
 
     const chartData = paged.map((repo) => {
-      const weight = parseFloat(String(repo.config?.weight ?? 0));
+      const weight = Number(repo.weight) || 0;
       const ossScore = repo.totalScore || 0;
       const prs = repo.totalPRs || 0;
       const contributors = repo.uniqueMiners?.size || 0;


ventura-oss:
Thanks for catching the zero-bar bug, @blinkeye-lcm — the diagnosis is right, but the fix doesn't quite land. `repo.weight` is `null` for every repo in the live `/dash/repos` data, so `Number(repo.weight) || 0` still evaluates to `0` and the bars stay flat. The value you want is `config.emissionShare` — that's what the chart's axis max (`maxWeight`) and the weight column/sort already read, so the bars need to match it:

```ts
const weight = parseFloat(String(repo.config?.emissionShare ?? 0)) || 0;
```

Could you switch to that field and rebase on the latest `test`? It's drifted behind since #1277 merged into the same file. (The red CI here is just a checkout-step hiccup, not your code.) Thanks!

e35ventura:
CI failing

PR #1258, opened 2026-05-21
fix(code-browser): hide file search when viewing a file

sunlight7777777:
The "Search files and folders" input was still rendered when a file
was open, even though there's no file list to filter at that point.
Conditionally render GoToFileSearch only in directory view.

### Screenshots
- **Directory view** — search input is appropriate here:
  _(attach Image 1: repo root showing folders/files)_

Image

- **File view (`.env.example`)** — search input should not be shown here:
  _(attach Image 2: opened file contents)_

Image

Fixes #1256

PR #1279, opened 2026-05-27
fix(DataTable): make default row hover visible on dark theme

### LINKED ISSUE #1278: [Bug] `DataTable` rows have invisible hover state on dark theme
Summary
The default row hover styling in the shared DataTable component uses theme.palette.surface.subtle as the hover background. On Gittensor's dark theme this color is visually indistinguishable from the surrounding table background, so the hover state fires but users can't see it. Rows that are clickable look non-interactive.
This affects every consumer of DataTable that uses onRowClick or getRowHref without passing a custom getRowSx override.

Steps to reproduce

Open the app and navigate to any repository detail page.
Click the Issues tab.
Move the cursor across rows in the issues table.

Screenshot
Before fix:

https://github.com/user-attachments/assets/66ef3a9c-5027-47c2-ab79-97d64a3a353c

After fix:

https://github.com/user-attachments/assets/a2909eae-11f3-4d0b-b1b9-fa38cd9c5ee8

Actual behavior

Tables that don't pass getRowSx (e.g. RepositoryIssuesTable) have no visible hover. The hover is being applied — DevTools confirms the background-color change — but the chosen color is too close to the page background to perceive.
Tables that do pass getRowSx (e.g. TopRepositoriesTable) look correct, but only because each consumer duplicates the same styling. This is a code-smell: every new DataTable consumer has to remember to add the override, or the table looks broken.

Root cause
In src/components/common/DataTable.tsx:
tsxconst clickableRowSx: SxProps<Theme> = (theme) => ({
  cursor: 'pointer',
  transition: 'background-color 0.2s',
  '&:hover': {
    backgroundColor: theme.palette.surface.subtle, // ← invisible on dark theme
  },
});
surface.subtle is approximately alpha(white, 0.02). Against a transparent table card on a near-black page background, the contrast difference is below the threshold of human perception.
The original intent of the code is clearly to show a hover state — cursor: 'pointer', transition, :hover background are all defined. The default just picked a color that doesn't deliver on that intent.

sunlight7777777:
## Summary

Fixes the invisible row hover state on `DataTable`. The default used
`surface.subtle`, which on our dark theme is indistinguishable from
the table background.

Closes #1278 

## Change

Single-file CSS change in `src/components/common/DataTable.tsx`:

- Hover background: `surface.subtle` → `border.subtle`
- Added `borderBottom: 1px solid surface.light` for row separation
- Widened `transition` from `background-color` to `all`

This matches the styling `TopRepositoriesTable` already uses via its
own `getRowSx` override.

## Affected surfaces

- ✅ `RepositoryIssuesTable` — was broken, now visibly hoverable
- ✅ Any other `DataTable` consumer relying on the default — same fix
- ⚪ `TopRepositoriesTable` — no visual change (its consumer-level
  override matches the new default)

## Testing

- [x] Repository → Issues tab: rows highlight on hover, cursor changes
      to pointer, click opens GitHub.
- [x] Repositories list view: hover behavior unchanged.
- [x] Lint passes.
- [x] Type check passes.

##Screenshot
Before Change:

https://github.com/user-attachments/assets/a4f74c90-9bf4-49d2-9038-b9bad6926b62

After Change:

https://github.com/user-attachments/assets/2d3b45e3-e8c3-4c3f-8691-ff1c5a7964cb

### DIFF
--- a/src/components/common/DataTable.tsx
+++ b/src/components/common/DataTable.tsx
@@ -120,9 +120,10 @@ const tableSx = {
 
 const clickableRowSx: SxProps<Theme> = (theme) => ({
   cursor: 'pointer',
-  transition: 'background-color 0.2s',
+  transition: 'all 0.2s',
+  borderBottom: `1px solid ${theme.palette.surface.light}`,
   '&:hover': {
-    backgroundColor: theme.palette.surface.subtle,
+    backgroundColor: theme.palette.border.subtle,
   },
 });
 


e35ventura (approved):

sunlight7777777:
My prs are merged so have only one open pr now but #1287  pr is still closed
I should create new pr for #1258 ?
@anderdc
@ventura-oss

PR #1319, opened 2026-06-04
fix(leaderboard): align sparkline totals with contributions column

### LINKED ISSUE #1318: Leaderboard sparkline totals do not match the CONTRIBUTIONS column
## Description

The 30 day activity sparkline on the miners leaderboard reports totals that disagree with the CONTRIBUTIONS column. The OSS track reads low and the discovery track reads high.

Root causes:

- OSS is built from the global recent commits feed (`/dash/commits`, capped at 500 rows). Once those rows are spread across every miner, the feed only covers one to two weeks, so merged PRs that are still inside the 30 day window get dropped.
- Discovery counts a solved issue from just the issue author plus the presence of a merged solving PR. It applies none of the validator classification gates and is not scoped to the scored repositories.

The CONTRIBUTIONS column is the validator scored aggregate, so the sparkline has to be built from the same data and the same gates for the totals to agree.

## Steps to Reproduce

1. Open the miners leaderboard.
2. Hover a miner row to expand the 30d activity tooltip.
3. Compare the OSS and Discovery totals in the tooltip against the OSS and Discovery values in the CONTRIBUTIONS column.

## Expected Behavior

The sparkline OSS total equals `totalMergedPrs` and the Discovery total equals `totalSolvedIssues` over the same window.

## Actual Behavior

OSS is undercounted and Discovery is overcounted. One observed row showed OSS 11 and Discovery 28 in the tooltip while the column reported OSS 14 and Discovery 2.

## Environment

- Browser: 
- OS: 

## Additional Context

The full scored pull request feed (`/prs`) already exposes the same set the column counts and includes the GitHub id for direct attribution. The mirror issues payload already carries the fields needed to apply the validator classification, so both tracks can be reconciled in the client with no backend change.

Dexterity104:
## Summary

The miners leaderboard 30 day sparkline now agrees with the CONTRIBUTIONS column on both tracks. All changes live in `src/components/leaderboard/useMinerActivityIndex.ts`.

OSS track:

- The merged PR series now reads the full scored pull request feed (`/prs`) instead of the global recent commits feed (`/dash/commits`, capped at 500 rows). The capped feed only covered one to two weeks once spread across every miner, so merged PRs that were still inside the window went missing and the total fell short of `totalMergedPrs`. The `/prs` feed is the same gated set the column counts and carries the GitHub id directly, so attribution no longer depends on the hotkey or login fallback.

Discovery track:

- The solved issue gate now mirrors the validator classification: closed as completed, solved by a merged PR that was not edited after merge, issue not transferred, and issue not edited after the solving PR merged. It is also scoped to the active scored repositories. The previous version counted any authored issue that had a merged solving PR, which overcounted against `totalSolvedIssues`.

The broad `total_solved_issues` semantics are matched on purpose. The token score valid gate and the one issue per PR dedupe are intentionally left out, since those are not what the column shows, and a code comment records this so the gate set is not "corrected" by mistake later. Branch eligibility is the only validator gate that cannot run client side, because the issue payload does not include the base and head branch names.

## Related Issues

Fixes #1318 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Screenshots

Before: 
before

After: 
after

## Checklist

- [x] `npm run build` passes

### DIFF
--- a/src/components/leaderboard/MinersLeaderTable.tsx
+++ b/src/components/leaderboard/MinersLeaderTable.tsx
@@ -2477,8 +2477,28 @@ const MinersLeaderTable: React.FC<MinersLeaderTableProps> = ({
 }) => {
   const navigate = useNavigate();
   const { isWatched, count: watchedCount } = useWatchlist('miners');
+  const activityGithubIds = useMemo(
+    () => miners.map((m) => m.githubId).filter(Boolean),
+    [miners],
+  );
+  // `/dash/commits` carries hotkey + author login but NOT githubId, so the
+  // OSS/merged series can't be attributed without this identity bridge.
+  const minerIdentities = useMemo(
+    () =>
+      miners.map((m) => ({
+        githubId: m.githubId,
+        hotkey: m.hotkey,
+        login: m.githubUsername,
+      })),
+    [miners],
+  );
   const { index: activityIndex, isLoading: isLoadingActivity } =
-    useMinerActivityIndex({ lookbackDays: 30, topReposLimit: 2 });
+    useMinerActivityIndex({
+      lookbackDays: 30,
+      topReposLimit: 2,
+      githubIds: activityGithubIds,
+      identities: minerIdentities,
+    });
 
   const [searchParams, setSearchParams] = useSearchParams();
   const initialViewMode = useMemo<LeaderboardViewMode>(() => {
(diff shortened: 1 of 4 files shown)


Dexterity104:
Hi @e35ventura 
Could you please check this PR?

anderdc:
The mismatch is real, but this fixes it by fetching every rendered miner's full issue history separately — roughly one mirror request per miner (~110) fired eagerly on every leaderboard load, growing with the roster. That per-miner fan-out is exactly the cost the bulk dashboard-issues endpoint exists to avoid, and it's too much to route a secondary sparkline track through. Closing.

PR #1370, opened 2026-08-27
feat(compute): /compute fleet page + per-hotkey compute miner status

anderdc:
## What

Makes the serving (compute) sub-mechanism visible to miners. Built against the das `/serving` contract with dev-only mocked data; the endpoints are not live yet.

- **`/compute`** — header + one-line pool explainer, KPI row (READY cards, pool share "x.xx % of 3.5 % cap", card-equivalents, est. α/day per full card + USD, last round + rounds in 24 h), sortable fleet `DataTable` (UID, hotkey copy, status pill, window `n/10 · mean`, tok/s, TTFT credit, capacity, settled, uptime 24 h, est. α/day, last miss w/ tooltip). Row click → detail. 404 → "No serving rounds recorded yet by this validator."
- **`/compute/miner?hotkey=…`** — identity header (UID, hotkey copy, `@username` → `/miners/details?githubId=` when present, model id), status card with plain-English "why" per status, KPI row, three ECharts line charts over the last 24 h of rounds (score, probe tok/s, TTFT credit) using the gittensor chart theme + `ChartEmptyPanel`, misses table.
- Every compute page carries the **validator snapshot footnote** ("Snapshot as observed by validator `<hotkey, truncated w/ copy>`. Other validators run their own audits; scores and payouts shown here are estimates, not consensus."). All payout numbers are labelled est.
- Nav: "compute" entry in the sidebar (next to dashboard — the sidebar has no Leaderboard entry since the one-page UI pass). Global search: an ss58-looking query (48 chars, leading `5`) gets a "Compute miner …" result linking to the detail page.
- `computePath()` / `computeMinerPath(hotkey)` / `looksLikeSs58Hotkey()` in `src/utils/paths.ts`.

## API layer

`src/api/ServingApi.ts` + `src/api/models/Serving.ts` — `useServingStatus`, `useServingMiners`, `useServingMinerDetail(hotkey, hours)`; 60 s refetch; numeric coercion (postgres `numeric` → string) like `MinerApi.ts`; `isNotFoundError` helper for the 404 empty states.

**Dev mock:** set `VITE_SERVING_MOCK=true` in `.env` (dev only; the fixture module is a lazy chunk and does not ship in the prod bundle).

## Contract assumptions

- `estAlphaPerCardDay` / `estUsdPerCardDay` are per *full* card-equivalent; `estAlphaPerDay` per miner is already scaled by settled score.
- `uid` and numeric fields may arrive as strings (TypeORM numeric) — coerced client-side.
[...]

PR #1371, opened 2026-08-28
compute pages: observed TTFT (ms) + decode tok/s, settled-score explanations

anderdc:
Compute page fixes from watching soak 6:

- **TTFT** column shows the validator-recorded milliseconds (e.g. `48 ms`) instead of the 0–1 credit; the credit moves to its own **Speed** column.
- **tok/s** is the observed decode rate on served traffic (the old `probeTps` was never populated after #1722). Both show for probation miners too.
- **Settled** and every other metric header has a tooltip; miner page KPIs / charts / footer rewritten to the current mechanism (round = window pass × speed credit × attested; settled = 12-round mean, what the miner is paid on). Stale `capacity (probe tok/s ÷ 180)` copy removed.

Needs das-gittensor #94 (`ttftMs` / `decodeTps`). Includes the earlier compute-page commits on this branch.

https://claude.ai/code/session_01Md8Vw3LJe91aNwrsje1UBY

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,3 +4,6 @@ VITE_REACT_APP_MIRROR_BASE_URL=https://mirror.gittensor.io/api/v1
 
 # Web3Forms — repository registration form delivery
 VITE_WEB3FORMS_ACCESS_KEY=
+
+# Dev only: mock the /serving (compute) endpoints until das ships them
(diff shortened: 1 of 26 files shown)


anderdc:
Superseded by the rebased PR (base commits already landed via #1370).

